PR #24, opened 2026-02-03
Improve DX based on second round of evaluation feedback

mike-north:
## Summary

Addresses issues identified in the second round of DX evaluation:

- **P1: CLI Error Messages** - Provides context-aware error messages that distinguish between "compiled file missing" and "no FormSpec exports found" to help users understand when they need to compile TypeScript first
- **P2: Document codegen Command** - Adds comprehensive documentation for the `codegen` command in CLI README
- **P3: Document Output Formats** - Explains the difference between FormSpec Native format (`ux_spec.json`) and JSON Forms format (`uiSchema`)
- **P4: Conditional Field Type Inference** - Fixes `InferFormSchema` so fields inside `when()` conditionals are correctly typed as optional

## Test plan

- [x] All 374 tests pass across workspace
- [x] tsd type-level tests verify conditional fields are optional
- [x] Build succeeds and API report updated

### DIFF
--- a/packages/dsl/src/inference.ts
+++ b/packages/dsl/src/inference.ts
@@ -101,7 +101,21 @@ export type ExtractFieldsFromArray<Elements> = Elements extends readonly [
 
 /**
  * Extracts fields that are NOT inside conditionals.
- * These fields are always visible and should be required.
+ * These fields are always visible and should be required in the inferred schema.
+ *
+ * @example
+ * ```typescript
+ * // Given a form with conditional and non-conditional fields:
+ * const form = formspec(
+ *   field.text("name"),           // non-conditional
+ *   field.number("age"),          // non-conditional
+ *   when(is("type", "business"),
+ *     field.text("company"),      // conditional (skipped)
+ *   ),
+ * );
+ *
+ * // ExtractNonConditionalFields extracts: TextField<"name"> | NumberField<"age">
+ * ```
  */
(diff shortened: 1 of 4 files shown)


packages/cli/README.md
```diff
@@ -42,6 +42,137 @@ formspec generate ./src/forms.ts MyClass -o ./generated
...
+- You're using the decorator DSL (not chain DSL)
+- You need type information that TypeScript erases at compile time
+
+> **Note:** The chain DSL (`@formspec/dsl`) doesn't require codegen because it preserves type information at runtime.
+
+## Output Formats
+
+FormSpec supports two output formats for UI schemas:
+
+### FormSpec Native Format (`ux_spec.json`)
```
mike-north: Why do we need this internal format? Can't we derive what we need from JSONSchema / the JSONForm's ui schema format?

packages/dsl/src/inference.ts
```diff
@@ -99,6 +99,46 @@ export type ExtractFieldsFromArray<Elements> = Elements extends readonly [
   ? ExtractFields<First> | ExtractFieldsFromArray<Rest>
   : never;
 
+/**
+ * Extracts fields that are NOT inside conditionals.
+ * These fields are always visible and should be required.
+ */
+export type ExtractNonConditionalFields<E> = E extends AnyField
```
mike-north: Add examples to the TSDoc comments for all of these types

PR #119, opened 2026-03-26
test(e2e): replace alias-chain gold masters with spec assertions

mike-north:
## Summary
- add a three-level alias-chain fixture with direct normative assertions
- replace remaining alias/shared-type gold-master expectations with structural schema and UI assertions
- verify shared `$defs` reuse and open-object defaults on referenced types

## Test plan
- pnpm run build
- pnpm run lint
- pnpm run typecheck
- pnpm --filter @formspec/e2e exec vitest run tests/alias-chain-3-level.test.ts tests/tsdoc-shared-types.test.ts tests/tsdoc-type-alias-pipeline.test.ts tests/tsdoc-inherited-constraints.test.ts
- $ai_review (APPROVE, no findings)

### DIFF
--- a/e2e/tests/tsdoc-shared-types.test.ts
+++ b/e2e/tests/tsdoc-shared-types.test.ts
@@ -51,6 +51,7 @@ describe("TSDoc Shared Types ($defs/$ref)", () => {
     expect(required).toContain("shippingAddress");
     expect(required).not.toContain("notes");
   });
+
   it("omits additionalProperties for the shared Address object by default", () => {
     const defs = schema["$defs"] as Record<string, Record<string, unknown>>;
     expect(defs["Address"]["additionalProperties"]).toBeUndefined();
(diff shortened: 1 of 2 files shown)


e2e/tests/type-mappings-comprehensive.test.ts
```diff
@@ -37,142 +37,71 @@ describe("Type Mappings — comprehensive", () => {
...
+    expect(properties["optionalNumber"]).toEqual({ type: "number" });
 
-  describe("String literal union → enum", () => {
-    // @see 003-json-schema-vocabulary.md §2.3: "string literal union → flat enum array"
-    it("stringLiteralUnion: 'a' | 'b' | 'c' → enum: [a, b, c]", () => {
-      expect(properties["stringLiteralUnion"]["enum"]).toEqual(["a", "b", "c"]);
-      // No type alongside flat enum (enum is self-constraining)
-      expect(properties["stringLiteralUnion"]["type"]).toBeUndefined();
-    });
+    const required = schema["required"] as string[];
```
mike-north: The four positive `required` assertions that were here (`stringField`, `numberField`, `booleanField`, `namedType`) were removed, leaving only the negative assertions. All four fields use `!` (definite assignment) in the fixture, so they should appear in the `required` array per the JSON Schema spec.

Without positive assertions, a regression that drops required fields from the output would go undetected by this test. Consider restoring them — they're spec-derived, not gold-master.

Issue #92, opened 2026-03-25
E2E: Diagnostic quality and error code coverage


mike-north:
## Context
The spec defines specific error codes (ERR-301, ERR-401, ERR-501, etc.) for various invalid inputs. Tests should verify every error code is produced correctly.

## Spec references
- 002-tsdoc-grammar.md §6 (error codes)

### Every error code from the spec
For each error code defined in spec §6, create a fixture that triggers it and verify:
- Non-zero exit code from CLI
- Error message contains the field name
- Error message contains the constraint name
- Severity (error vs warning) is correct

### Error codes to cover
- TYPE_MISMATCH (numeric constraint on string, etc.)
- CONTRADICTING_CONSTRAINTS (min > max)
- UNKNOWN_PATH_TARGET (nonexistent subfield)
- CONSTRAINT_BROADENING (field relaxes inherited constraint)
- UNKNOWN_EXTENSION (unregistered custom constraint)

### Multiple errors in one file
- Fixture with 3+ errors → all reported, not just the first

### Diagnostic source location
- Error diagnostics reference correct file/line/column

## Implementation notes
- Work off `main` branch
- Create worktree: `git worktree add .worktrees/diagnostic-tests origin/main`
- Branch: `test/diagnostic-quality`
- Use `--validate-only` CLI flag for error-only tests
- Author: `Mike North <michael.l.north@gmail.com>`

PR #121, opened 2026-03-26
fix(build): validate invalid path targets semantically (#92)

mike-north:
## Summary
- validate path-targeted constraints against the resolved nested target type
- emit `TYPE_MISMATCH` for incompatible nested targets and non-traversable array-item paths
- add focused unit and E2E coverage for invalid path-target diagnostics

## Test plan
- pnpm run build
- pnpm run lint
- pnpm run typecheck
- pnpm --filter @formspec/build exec vitest run src/__tests__/constraint-validator.test.ts
- pnpm --filter @formspec/e2e exec vitest run tests/tsdoc-path-target.test.ts
- $ai_review (no actionable findings; one dismissed false positive against a non-branch file)

### DIFF
--- a/packages/build/src/__tests__/constraint-validator.test.ts
+++ b/packages/build/src/__tests__/constraint-validator.test.ts
@@ -1005,6 +1005,38 @@ describe("validateIR", () => {
       expect(result.diagnostics[0]?.message).toContain("only valid on number fields");
     });
 
+    it("emits TYPE_MISMATCH when a path target references an unknown nested property", () => {
+      const objectType: ObjectTypeNode = {
+        kind: "object",
+        properties: [
+          {
+            name: "currency",
+            type: STRING_TYPE,
+            optional: false,
+            constraints: [],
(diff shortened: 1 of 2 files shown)


packages/build/src/__tests__/constraint-validator.test.ts
```diff
@@ -973,6 +973,62 @@ describe("validateIR", () => {
       expect(result.diagnostics[0]?.message).toContain("cannot be traversed");
     });
 
+    it("emits TYPE_MISMATCH when a path target resolves to an incompatible nested type", () => {
```
mike-north: The two new unit tests cover "resolved type is incompatible" and "unresolvable (primitive array items)". There's no unit test for the \`missing-property\` resolution path — when a path segment references a property that doesn't exist on the object type.

The e2e test covers the full pipeline but a targeted unit test for \`missing-property\` would verify the error message format (\`references unknown path segment\`) directly.

PR #122, opened 2026-03-26
feat(build): report constraint broadening semantically (#92)

mike-north:
## Summary
- emit semantic `CONSTRAINT_BROADENING` diagnostics for less restrictive follow-up constraints
- add validator coverage for broadening, including mixed numeric bound cases
- enable the corresponding e2e diagnostics coverage

## Verification
- pnpm --filter @formspec/build exec vitest run src/__tests__/constraint-validator.test.ts
- pnpm --filter @formspec/e2e exec vitest run tests/diagnostics-supported.test.ts
- pnpm run lint
- pnpm run typecheck
- pnpm run build

### DIFF
--- a/packages/build/src/__tests__/constraint-validator.test.ts
+++ b/packages/build/src/__tests__/constraint-validator.test.ts
@@ -492,8 +492,10 @@ describe("validateIR", () => {
       const ir = makeIR([
         makeField("quantity", NUMBER_TYPE, [minConstraint(10, 1), minConstraint(12, 2)]),
       ]);
+      const result = validateIR(ir);
 
-      expect(validateIR(ir).valid).toBe(true);
+      expect(result.valid).toBe(true);
+      expect(result.diagnostics).toHaveLength(0);
     });
 
     it("emits CONSTRAINT_BROADENING when a later minimum broadens an earlier exclusiveMinimum at the same value", () => {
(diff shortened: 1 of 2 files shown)


packages/build/src/validate/constraint-validator.ts
```diff
@@ -154,6 +159,188 @@ function findAllowedMembers(
...
+  | "minimum"
+  | "exclusiveMinimum"
+  | "minLength"
+  | "minItems"
+  | "maximum"
+  | "exclusiveMaximum"
+  | "maxLength"
+  | "maxItems";
+
+type OrderedBoundConstraint = Extract<ConstraintNode, { readonly constraintKind: OrderedBoundKind }>;
```
mike-north: This `Extract` doesn't produce the intended type. Because `NumericConstraintNode["constraintKind"]` includes `"multipleOf"` (outside `OrderedBoundKind`), the entire `NumericConstraintNode` member fails the `extends` check and gets excluded. The resolved type is effectively just `LengthConstraintNode`.

Runtime behavior is correct because `isOrderedBoundConstraint` narrows correctly at every call site. However, the type system can't verify the numeric-lower/numeric-upper branches in `compareConstraintStrength` — a future refactor could silently break the numeric comparison logic without a type error.

Consider replacing with a structural type that captures what `compareConstraintStrength` actually needs:

```suggestion
type OrderedBoundConstraint = {
  readonly kind: "constraint";
  readonly constraintKind: OrderedBoundKind;
  readonly value: number;
  readonly path?: { readonly segments: readonly string[] };
  readonly provenance: Provenance;
};
```

packages/build/src/validate/constraint-validator.ts
```diff
@@ -31,9 +31,8 @@ import type { ExtensionRegistry } from "../extensions/index.js";
...
  * A structured diagnostic produced by constraint validation.
  *
- * The `code` follows the format: `{VENDOR}-{CATEGORY}-{NNN}`.
- * - VENDOR defaults to "FORMSPEC" (configurable via `vendorPrefix`).
- * - Categories: CONTRADICTION, TYPE_MISMATCH, UNKNOWN_EXTENSION
+ * The `code` is a stable, machine-readable semantic identifier.
+ * Examples: `CONTRADICTING_CONSTRAINTS`, `TYPE_MISMATCH`, `UNKNOWN_EXTENSION`.
  */
 export interface ValidationDiagnostic {
   readonly code: string;
```
mike-north: Now that there are four known diagnostic codes, consider typing this as a string literal union so consumers can pattern-match exhaustively:

```suggestion
  readonly code:
    | "CONTRADICTING_CONSTRAINTS"
    | "TYPE_MISMATCH"
    | "UNKNOWN_EXTENSION"
    | "CONSTRAINT_BROADENING";
```

Could be a follow-up if backward compatibility is a concern.

packages/build/src/__tests__/constraint-validator.test.ts
```diff
@@ -469,39 +469,95 @@ describe("validateIR", () => {
...
+      expect(diag?.primaryLocation).toEqual(prov(2, "minimum"));
+      expect(diag?.relatedLocations).toEqual([prov(1, "minimum")]);
+    });
+
+    it("does not emit when a later minimum narrows the earlier bound", () => {
+      const ir = makeIR([
+        makeField("quantity", NUMBER_TYPE, [minConstraint(10, 1), minConstraint(12, 2)]),
+      ]);
+
+      expect(validateIR(ir).valid).toBe(true);
```
mike-north: Nit: Since `valid` is defined as "no error-severity diagnostics", a hypothetical regression that emitted `CONSTRAINT_BROADENING` as a warning would still pass this test. Consider asserting `diagnostics.toHaveLength(0)` alongside `valid`:

```suggestion
      expect(validateIR(ir).valid).toBe(true);
```

Could become:

```typescript
const result = validateIR(ir);
expect(result.valid).toBe(true);
expect(result.diagnostics).toHaveLength(0);
```

Issue #89, opened 2026-03-25
E2E: Comprehensive test matrix implementation (13 fixture groups)


mike-north:
## Context
A spec-first test matrix was generated at `scratch/e2e-test-matrix.md` (1725 lines, 13 fixture groups). This issue tracks implementing all remaining fixture groups not covered by other issues/PRs.

## Test matrix location
`scratch/e2e-test-matrix.md` in the repo root. This is the source of truth — every fixture, every expected JSON Schema output, and every assertion is documented there with spec section references.

## Fixture groups to implement

| # | Group | Fixtures | Status |
|---|-------|----------|--------|
| 1 | Annotation tags (@displayName, @description, @placeholder, @deprecated, @defaultValue) | 3 fixtures | Needed |
| 2 | Numeric constraints (all permutations) | 1 fixture, 13 fields | Needed |
| 3 | String constraints (all permutations) | 1 fixture, 13 fields | Needed |
| 4 | Array constraints | 1 fixture, 8 fields | Needed |
| 5 | Type mappings (all TS → JSON Schema) | 1 fixture, 14 fields | Needed |
| 6 | 3-level alias chains | 2 fixtures | Partially exists (PR #84) |
| 7 | Path-target expanded | 1 fixture | Partially exists |
| 8 | Enum display names (:member syntax) | 1 fixture | Needed |
| 9 | Parity (TSDoc ↔ chain DSL) | 2 fixtures | Partially exists (PR #82) |
| 10 | Error cases | 4 fixtures | Partially exists (PR #84) |
| 11 | Spec 006 canonical fixtures (USDCents, PlanStatus, Address) | 3 fixtures | Needed |
| 12 | Exclusive bound edge cases | 1 fixture | Needed |
| 13 | @const tag | 1 fixture | Needed |

## CRITICAL rules
- **Every assertion derived from the spec** — write expected values by hand
- **NO gold-masters** — no `loadExpected()`, no `e2e/expected/` files, no `toMatchSnapshot()`
- **Tests that fail = bugs found** → mark `.skip("BUG: description")` with spec ref
- **Spec references on every assertion** as comments

## Implementation notes
- Work in `.worktrees/` at repo root
- Read `scratch/e2e-test-matrix.md` for exact fixture code and expected output
- Author: `Mike North <michael.l.north@gmail.com>`

PR #123, opened 2026-03-26
test(build): add remaining static parity fixtures (#89)

mike-north:
## Summary
- add the remaining shared static-surface parity fixtures for `plan-status` and `usd-cents`
- keep parity assertions hand-written and spec-driven rather than relying on gold masters
- limit the slice to static shared-surface behavior only

## Verification
- pnpm install --frozen-lockfile
- pnpm run build
- pnpm --filter @formspec/build exec vitest run src/__tests__/parity/parity.test.ts
- pnpm run lint
- pnpm run typecheck

### DIFF
--- a/packages/build/src/__tests__/parity/parity.test.ts
+++ b/packages/build/src/__tests__/parity/parity.test.ts
@@ -60,12 +60,12 @@ function canonicalizeFixtureClass(fixtureFile: string, className: string) {
   return canonicalizeTSDoc(analysis, { file: fixtureFile });
 }
 
-type ParityFixture = {
+interface ParityFixture {
   readonly name: string;
   readonly chainForm: Parameters<typeof canonicalizeChainDSL>[0];
   readonly expectedIR: ReturnType<typeof stripProvenance>;
   readonly className: string;
-};
+}
 
 const parityFixtures: readonly ParityFixture[] = [
   {


packages/build/src/analyzer/class-analyzer.ts
```diff
@@ -645,7 +645,7 @@ function resolveObjectType(
 
   // Circular reference guard
   if (visiting.has(type)) {
-    return { kind: "object", properties: [], additionalProperties: true };
+    return { kind: "object", properties: [], additionalProperties: false };
   }
   visiting.add(type);
```
mike-north: This fix changes the circular reference sentinel from `additionalProperties: true` to `false`. The normal object resolution path (line 690) still uses `true`, and the IR type docs say ordinary objects default to `true`.

The distinction is arguably correct (a circular placeholder shouldn't claim to accept arbitrary properties), but the existing test for self-referential types (`ir-analyzer.test.ts:550`) only asserts the result doesn't stack overflow and has a valid `kind` -- it doesn't verify `additionalProperties` on the sentinel.

Consider adding an assertion to that test:

```suggestion
    return { kind: "object", properties: [], additionalProperties: false };
```

And in `ir-analyzer.test.ts`, after the existing assertions for `selfRefRecord`:

```typescript
// Circular reference sentinel should not claim to accept additional properties
if (field?.type.kind === "object") {
  expect(field.type.additionalProperties).toBe(false);
}
```

This documents the intent and prevents accidental regression.

packages/build/src/__tests__/parity/parity.test.ts
```diff
@@ -149,6 +153,68 @@ describe("product-config parity", () => {
   });
 });
 
+// =============================================================================
+// plan-status parity
+// =============================================================================
+
+describe("plan-status parity", () => {
```
mike-north: With this PR there are now five structurally identical parity `describe` blocks — each differs only in the fixture name, imported form, expected IR, and TSDoc class name. A data-driven approach would reduce ~150 lines to ~25 and make adding future fixtures a one-line table entry:

```typescript
const parityFixtures = [
  { name: "address", form: addressForm, expected: addressExpected, className: "AddressForm" },
  { name: "user-registration", form: userRegistrationForm, expected: userRegistrationExpected, className: "UserRegistrationForm" },
  { name: "product-config", form: productConfigForm, expected: productConfigExpected, className: "ProductConfigForm" },
  { name: "plan-status", form: planStatusForm, expected: planStatusExpected, className: "SubscriptionForm" },
  { name: "usd-cents", form: usdCentsForm, expected: usdCentsExpected, className: "LineItemForm" },
];

for (const { name, form, expected, className } of parityFixtures) {
  describe(`${name} parity`, () => {
    it("chain DSL produces expected IR", () => {
      const ir = canonicalizeChainDSL(form);
      expect(stripProvenance(ir)).toEqual(expected);
    });

    it("TSDoc produces expected IR", () => {
      const fixturePath = nodePath.join(fixturesDir, name, "tsdoc.ts");
      const ir = canonicalizeFixtureClass(fixturePath, className);
      expect(stripProvenance(ir)).toEqual(expected);
    });

    it("both surfaces produce identical IR", () => {
      const chainIR = canonicalizeChainDSL(form);
      const fixturePath = nodePath.join(fixturesDir, name, "tsdoc.ts");
      const tsdocIR = canonicalizeFixtureClass(fixturePath, className);
      expect(compareIR(chainIR, tsdocIR)).toEqual([]);
    });
  });
}
```

This eliminates the copy-paste pattern and makes each new fixture a one-line addition to the table. The `compareIR utility tests` block at the bottom stays as-is since it tests the utility itself.

packages/build/src/__tests__/parity/fixtures/usd-cents/expected-ir.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import type { ProvenanceFreeFormIR } from "../../utils.js";
+
+export const expectedIR: ProvenanceFreeFormIR = {
+  kind: "form-ir",
+  irVersion: "0.1.0",
+  elements: [
+    {
+      kind: "field",
+      name: "unitPrice",
```
mike-north: Nit: the expected IR asserts a specific constraint order (`minimum` before `multipleOf`) for both fields. Is this ordering contractually guaranteed by the analyzer (e.g., spec-defined or alphabetical), or is it incidental (depends on JSDoc tag parse order or Map iteration)?

If it's contractual, a brief comment here would clarify that for future contributors. If it's incidental, the parity test utilities could sort constraints by `constraintKind` before comparison to avoid brittleness.

Issue #25, opened 2026-02-05
Add dry-run mode to CLI


mike-north:
## Summary

Add a `--dry-run` flag to the `formspec generate` command that shows what the CLI detects without writing any files.

## Use Case

When debugging or exploring, users want to see:
- Which decorated classes were detected
- What fields/methods were found on each class
- Which FormSpec exports were discovered
- What schemas would be generated

Without actually writing any output files.

## Suggested Implementation

```bash
formspec generate ./src/forms.ts MyClass --dry-run
```

Output would show:
```
Detected class: MyClass
  Fields: name (string), age (number), status (enum)
  Instance methods: activate(params), deactivate()
  Static methods: create(data)

Would generate:
  - generated/MyClass/schema.json
  - generated/MyClass/ux_spec.json
  - generated/MyClass/instance_methods/activate/params.schema.json
  ...
```

## Origin

From PR #24 review feedback.

PR #124, opened 2026-03-26
feat(cli): add dry-run mode to generate (#25)

mike-north:
## Summary
- add `--dry-run` to `formspec generate` so the CLI reports planned outputs without writing files
- share output planning between real writes and dry runs so `schema.json`, `ui_schema.json`, `params.ui_schema.json`, and optional `*.ir.json` stay aligned
- add subprocess coverage for dry-run behavior and document the new flag in the CLI README

## Test plan
- [x] pnpm install
- [x] pnpm run build
- [x] pnpm run lint
- [x] pnpm run typecheck
- [x] pnpm --filter @formspec/cli exec vitest run src/__tests__/dry-run-subprocess.test.ts src/__tests__/emit-ir-validate-only.test.ts src/__tests__/integration.test.ts
- [x] pnpm --filter @formspec/cli run test

### DIFF
--- a/packages/cli/src/output/writer.ts
+++ b/packages/cli/src/output/writer.ts
@@ -52,8 +52,8 @@ interface ClassSchemaFilePaths {
 
 interface MethodSchemaFilePaths {
   methodDir: string;
-  paramsSchemaPath?: string;
-  paramsUiSchemaPath?: string;
+  paramsSchemaPath: string | undefined;
+  paramsUiSchemaPath: string | undefined;
   returnTypePath: string;
 }
 


packages/cli/src/__tests__/dry-run-subprocess.test.ts
```diff
@@ -0,0 +1,213 @@
+import { afterAll, describe, expect, it } from "vitest";
+import * as fs from "node:fs";
+import * as path from "node:path";
+import { spawnSync } from "node:child_process";
+
+const packageDir = path.resolve(__dirname, "..", "..");
+const cliPath = path.join(packageDir, "dist", "index.js");
+const tempRoot = path.join(__dirname, "__dry_run_tmp__");
```
mike-north: Placing temp directories inside the source tree risks accidental commits if the `afterAll` cleanup doesn't run (test crash, timeout, Ctrl-C). Every other test file in the repo uses `os.tmpdir()` for this.

```suggestion
const tempRoot = path.join(os.tmpdir(), "formspec-cli-dry-run-test");
```

(This also requires adding `import * as os from "node:os";` to the imports.)

packages/cli/src/__tests__/dry-run-subprocess.test.ts
```diff
@@ -0,0 +1,213 @@
...
+      'import { formspec, field, type InferFormSchema } from "@formspec/dsl";',
+      "",
+      "export const ActivateParams = formspec(",
+      '  field.number("amount", { label: "Amount", required: true })',
+      ");",
+      "",
+      "export class BillingPlan {",
+      "  total!: number;",
+      "",
+      "  activate(params: InferFormSchema<typeof ActivateParams>): { success: boolean } {",
```
mike-north: Good coverage of the happy paths. A few gaps worth considering:

1. **\`--dry-run\` without \`-o\`** — all 4 tests pass \`-o <dir>\`. When omitted, \`outDir\` defaults to \`"./generated"\` (relative). The dry-run output would show relative paths — worth verifying.
2. **\`--dry-run --validate-only\` when validation fails** — the tests cover the passing case, but not the case where constraint violations trigger exit code 1. That interaction is worth locking in.
3. **Simple class-only dry-run** (without \`--validate-only\` or \`--emit-ir\`) — tests 1 uses a mixed fixture, test 4 uses \`--validate-only\`. The simplest use case (just \`--dry-run\` on a class) is untested.

Issue #91, opened 2026-03-25
E2E: Extension API constraint tests


mike-north:
## Context
The constraint system is extensible via `createExtensionRegistry` and `defineConstraint`. E2E tests need to verify the extension API works end-to-end.

### Custom constraint registration
- Register a custom constraint via the extension API
- Apply it to a field in a fixture
- Verify it appears in the generated JSON Schema output

### Custom constraint on wrong type
- Register a constraint that only applies to `number` fields
- Apply it to a `string` field
- Verify a diagnostic is emitted

### Multiple extensions
- Register constraints from two different vendor namespaces
- Verify both appear correctly in output

### Extension constraint + built-in constraint on same field
- Field has both `@minimum 0` and a custom extension constraint
- Both should appear in output

## Implementation notes
- Work off `main` branch
- Create worktree: `git worktree add .worktrees/extension-tests origin/main`
- Branch: `test/extension-api-e2e`
- Author: `Mike North <michael.l.north@gmail.com>`

PR #125, opened 2026-03-26
feat(build): support extensions in public schema generation (#91)

mike-north:
## Summary
- expose extension registry support on the public `@formspec/build` surface, including `buildFormSchemas`, `generateJsonSchema`, `generateJsonSchemaFromIR`, and `writeSchemas`
- thread `extensionRegistry` and `vendorPrefix` through JSON Schema generation so custom types, constraints, and annotations emit through the public generator path
- add integration-style tests that exercise extension-backed schema generation from the public build exports and update the API report and changeset

## Test plan
- [x] pnpm install --frozen-lockfile
- [x] pnpm --filter @formspec/build run build
- [x] pnpm --filter @formspec/build run typecheck
- [x] pnpm --filter @formspec/build exec vitest run src/__tests__/extension-runtime.integration.test.ts src/__tests__/extension-api.test.ts src/__tests__/ir-json-schema-generator.test.ts
- [x] pnpm exec eslint packages/build/src/index.ts packages/build/src/browser.ts packages/build/src/json-schema/generator.ts packages/build/src/json-schema/ir-generator.ts packages/build/src/__tests__/extension-runtime.integration.test.ts

### DIFF
--- a/packages/build/src/__tests__/extension-runtime.integration.test.ts
+++ b/packages/build/src/__tests__/extension-runtime.integration.test.ts
@@ -3,6 +3,7 @@ import {
   buildFormSchemas,
   createExtensionRegistry,
   generateJsonSchemaFromIR,
+  writeSchemas,
   type GenerateJsonSchemaFromIROptions,
 } from "../index.js";
 import {
@@ -20,6 +21,9 @@ import {
   type Provenance,
 } from "@formspec/core";
 import { field, formspec } from "@formspec/dsl";
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
 
 const PROVENANCE: Provenance = {
   surface: "extension",
@@ -71,7 +75,7 @@ const moneyType = defineCustomType({
 const currencyConstraint = defineConstraint({
   constraintName: "Currency",
   compositionRule: "override",
-  applicableTypes: ["custom"],
+  applicableTypes: ["primitive", "custom"],
   toJsonSchema: (payload, vendorPrefix) => ({
     [`${vendorPrefix}-currency`]: payload,
   }),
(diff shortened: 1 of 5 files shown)


packages/build/src/json-schema/ir-generator.ts
```diff
@@ -682,3 +710,48 @@ function applyAnnotations(schema: JsonSchema2020, annotations: readonly Annotati
...
+
+function generateCustomType(type: CustomTypeNode, ctx: GeneratorContext): JsonSchema2020 {
+  const registration = ctx.extensionRegistry?.findType(type.typeId);
+  if (registration === undefined) {
+    throw new Error(
+      `Cannot generate JSON Schema for custom type "${type.typeId}" without a matching extension registration`
+    );
+  }
+
+  return registration.toJsonSchema(type.payload, ctx.vendorPrefix) as JsonSchema2020;
```
mike-north: `CustomTypeRegistration.toJsonSchema` returns `Record<string, unknown>` from `@formspec/core`. The cast to `JsonSchema2020` is unchecked — an extension could return any object shape and it would flow through as the *entire* schema for the field.

Unlike constraints and annotations (which merge onto an existing schema via `Object.assign`), custom types use the extension return value wholesale. A buggy extension returning `{ type: 42 }` would produce invalid JSON Schema with a confusing error far from the source.

At minimum, document the trust boundary:

```suggestion
  // Trust boundary: extensions are responsible for returning valid JSON Schema.
  // Core returns Record<string, unknown>; we cast because JsonSchema2020
  // cannot be a dependency of @formspec/core.
  return registration.toJsonSchema(type.payload, ctx.vendorPrefix) as JsonSchema2020;
```

packages/build/src/__tests__/extension-runtime.integration.test.ts
```diff
@@ -0,0 +1,236 @@
...
+      }
+    );
+
+    expect(schema.properties?.["amount"]).toEqual({
+      type: "string",
+      "x-stripe-money-scale": 2,
+    });
+  });
+
+  it("fails loudly when a custom type is generated without a matching extension registration", () => {
```
mike-north: This test covers the throw for custom *types* without a registry entry, but the analogous throws for custom *constraints* and custom *annotations* are untested. All three code paths (ir-generator.ts lines 718, 731, 745) have identical throw patterns — if a constraint or annotation throw regressed to a silent no-op, vendor properties would be silently dropped with no test failure.

Consider adding parallel tests:

```typescript
it("fails loudly when a custom constraint is generated without a matching extension registration", () => {
  expect(() =>
    generateJsonSchemaFromIR(
      makeIR([makeField("currencyCode", STRING_TYPE, [currencyConstraintNode("USD")])])
    )
  ).toThrow(/custom constraint.*without a matching extension registration/);
});
```

packages/build/src/index.ts
```diff
@@ -120,6 +131,11 @@ export interface BuildResult {
   readonly uiSchema: UISchema;
 }
 
+/**
+ * Options for building schemas from a FormSpec.
+ */
+export type BuildFormSchemasOptions = GenerateJsonSchemaOptions;
```
mike-north: \`BuildFormSchemasOptions\` → \`GenerateJsonSchemaOptions\` → \`GenerateJsonSchemaFromIROptions\` creates a three-level indirection where all three are structurally identical. Consumers must chase two aliases to discover the actual fields (\`extensionRegistry\` and \`vendorPrefix\`).

If future divergence is planned (e.g., canonicalization-specific options on the Chain DSL surface), a comment documenting that intent would help. Otherwise, consider collapsing to two levels.

```suggestion
/**
 * Options for building schemas from a FormSpec.
 *
 * Currently identical to {@link GenerateJsonSchemaFromIROptions}.
 * Defined separately so the Chain DSL surface can extend it independently
 * in the future (e.g., canonicalization-specific options).
 */
export type BuildFormSchemasOptions = GenerateJsonSchemaOptions;
```

PR #129, opened 2026-03-26
feat(build): emit literal and array constraint metadata

mike-north:
## Summary
- emit TSDoc literal and array constraint metadata into JSON Schema, including `@const`, `@format`, and `@uniqueItems`
- apply string item constraints and format annotations to primitive array item schemas where required by the spec
- add explicit end-to-end `@const` coverage and enable the remaining skipped string and array constraint tests

## Test plan
- pnpm --filter @formspec/build run build
- pnpm --filter @formspec/cli run build
- pnpm --filter @formspec/e2e exec vitest run tests/const-constraints.test.ts tests/string-constraints-comprehensive.test.ts tests/array-constraints-comprehensive.test.ts
- pnpm exec eslint e2e/tests/const-constraints.test.ts e2e/tests/string-constraints-comprehensive.test.ts e2e/tests/array-constraints-comprehensive.test.ts packages/build/src/__tests__/ir-jsdoc-constraints.test.ts packages/build/src/__tests__/parity/utils.ts packages/build/src/analyzer/tsdoc-parser.ts packages/build/src/json-schema/ir-generator.ts packages/build/src/validate/constraint-validator.ts packages/core/src/__tests__/ir.test-d.ts packages/core/src/index.ts packages/core/src/types/constraint-definitions.ts packages/core/src/types/index.ts packages/core/src/types/ir.ts

### DIFF
--- a/packages/build/src/analyzer/tsdoc-parser.ts
+++ b/packages/build/src/analyzer/tsdoc-parser.ts
@@ -445,6 +445,10 @@ function parseConstraintValue(
           provenance,
         };
       } catch {
+        // Bare strings like `@const USD` are accepted as a shorthand for
+        // string-valued const constraints. Non-string malformed JSON still
+        // gets rejected later during IR validation if the field type is
+        // incompatible with this fallback string value.
         return {
           kind: "constraint",
           constraintKind: "const",
--- a/e2e/tests/const-constraints.test.ts
+++ b/e2e/tests/const-constraints.test.ts
@@ -54,6 +54,6 @@ describe("Const Constraints", () => {
 
   it("all fields are required", () => {
     const required = schema["required"] as string[];
-    expect(required).toEqual(expect.arrayContaining(["currency", "statusCode", "enabled"]));
+    expect(required).toEqual(["currency", "statusCode", "enabled"]);
   });
 });
--- a/packages/build/src/__tests__/constraint-validator.test.ts
+++ b/packages/build/src/__tests__/constraint-validator.test.ts
@@ -719,8 +719,37 @@ describe("validateIR", () => {
   });
 
   describe("const constraint validation", () => {
+    it("accepts a primitive const value on a matching primitive field", () => {
+      const ir = makeIR([makeField("currency", STRING_TYPE, [constConstraint("USD", 1)])]);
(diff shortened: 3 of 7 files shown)


packages/build/src/analyzer/tsdoc-parser.ts
```diff
@@ -400,7 +411,46 @@ function parseConstraintValue(
...
+      provenance,
+    };
+  }
+
   if (expectedType === "json") {
+    if (tagName === "const") {
+      const text = effectiveText.trim();
+      if (text === "") return null;
+
+      try {
```
mike-north: The \`JSON.parse(text) as JsonValue\` cast is acceptable here. \`JSON.parse\` returns \`any\`, and the \`JsonValue\` type (\`null | boolean | number | string | readonly JsonValue[] | { readonly [key: string]: JsonValue }\`) covers the full range of JSON-representable values. Since \`JSON.parse\` can only produce JSON-representable values (by definition), the assertion is sound.

The fallback branch (catch clause, lines 443-451) treats unparseable text as a bare string literal, which makes sense for the \`@const USD\` shorthand (unquoted string). No issue here, just noting the design decision is reasonable.

packages/build/src/analyzer/tsdoc-parser.ts
```diff
@@ -400,7 +411,50 @@ function parseConstraintValue(
...
+          constraintKind: "const",
+          value: parsed,
+          ...(path && { path }),
+          provenance,
+        };
+      } catch {
+        return {
+          kind: "constraint",
+          constraintKind: "const",
+          value: trimmedText,
```
mike-north: When `JSON.parse` fails for a `@const` tag, the raw text is silently used as a string value. This creates an asymmetry: `@const true` produces `value: true` (boolean via JSON.parse), but `@const tru` silently produces `value: "tru"` (string) with no diagnostic. Similarly, `@const {bad}` becomes `value: "{bad}"`.

Every other constraint branch returns `null` on parse failure. If bare-string support is intentional (e.g., `@const USD` as shorthand for `@const "USD"`), it should be documented and tested. If not:

```suggestion
      } catch {
        return null;
      }
```

PR #273, opened 2026-04-16
Add brand field to CustomTypeRegistration for structural type detection

mike-north:
## Summary

Phase 2 of the `tsTypeNames` deprecation roadmap. Extensions can now register a `brand` string on custom types for structural detection via unique symbol brands, which is more reliable than name-matching (works with import aliases, prevents name collisions).

### Changes

- `CustomTypeRegistration` gains optional `brand?: string` field
- `ExtensionRegistry` builds a `brandMap` and exposes `findTypeByBrand`
- `resolveBrandedCustomType` in the type resolver checks intersection type properties for registered brands
- Shared `findBrandIdentifier` helper extracts brand identifiers (used by both builtin Integer detection and extension brand detection)
- Reserved brand check: `__integerBrand` cannot be registered by extensions
- 11 new tests covering registry lookup, duplicates, reserved brand, integration with real TS programs, alias resilience, and fallback

### Deprecation roadmap

| Phase | Status |
|---|---|
| 1. Builtin Integer via `__integerBrand` | Done (PR #272) |
| **2. `brand` field on `CustomTypeRegistration`** | **This PR** |
| 3. `defineCustomType<T>()` type parameter extraction | Planned |
| 4. Remove `tsTypeNames` | Major version |

## Test plan

- [x] 676 build tests pass (11 new)
- [x] All e2e tests pass (266)
- [x] Typecheck clean

PR #275, opened 2026-04-16
Add defineCustomType<T>() type parameter extraction for symbol-based detection

mike-north:
## Summary

Phase 3 of the `tsTypeNames` deprecation roadmap. When a config file uses `defineCustomType<T>()` with type parameters, the build pipeline walks the config AST, extracts the type argument, and resolves it to a `ts.Symbol` for O(1) identity-based lookup during field analysis. This is the most precise detection mechanism — immune to import aliases, namespace imports, renamed imports, and name collisions.

### Changes

- `defineCustomType` gains generic `<T = unknown>` parameter (erased at runtime, preserved in AST)
- `buildSymbolMapFromConfig` walks config AST using checker-based symbol resolution (handles `core.defineCustomType<T>()`, renamed imports, and re-exports)
- `ExtensionRegistry` gains `findTypeBySymbol`; `MutableExtensionRegistry` separates the `setSymbolMap` mutation from the read-only public interface
- `ExtensionTypeLookupResult` as shared return type for all `findTypeBy*` methods
- `resolveSymbolBasedCustomType` in the type resolver checks symbols after name-based, before brand-based
- `createProgramContext` accepts `additionalFiles` to include the config file in the TS program
- Batch mode hoists symbol map build out of per-target loop
- `tsTypeNames` marked `@deprecated`
- 15 new tests covering happy path, edge cases, namespace imports, priority ordering

### Deprecation roadmap

| Phase | Status |
|---|---|
| 1. Builtin Integer via `__integerBrand` | Done (#272) |
| 2. `brand` field on `CustomTypeRegistration` | Done (#273) |
| **3. `defineCustomType<T>()` type parameter extraction** | **This PR** |
| 4. Remove `tsTypeNames` | Major version |

## Test plan

- [x] 694 build tests pass (15 new)
- [x] All e2e tests pass (266)
- [x] Typecheck clean
- [x] Format clean


mike-north:
Reviewed PR #275 in an isolated worktree. I found the following issues.

1. [P1] `generateSchemasBatch()` seeds the symbol map once for the whole batch from the first `ProgramContext` only.
`packages/build/src/generators/class-schema.ts:624`
`ts.Symbol` identity is scoped to a specific `Program`, so every later target analyzed with a different `Program` will miss `findTypeBySymbol`. In practice, any batch using `configPath`-driven symbol detection will silently regress on every target after the first.

2. [P1] The new `configPath` contract is not actually honored across the full public API surface.
Files: `packages/build/src/generators/class-schema.ts:325`, `packages/build/src/generators/class-schema.ts:597`, `packages/build/src/generators/class-schema.ts:750`
`generateSchemasFromClass()` still builds a program from `filePath` only and never seeds the symbol map, and the `*FromProgram` APIs can only seed successfully if the caller already included the config file in their `Program`. That means the advertised `defineCustomType<T>()` path is not consistently available through the public entry points that now document it.

3. [P2] The deprecated `extensionRegistry` path now assumes mutability that the public type does not require.
Files: `packages/build/src/generators/class-schema.ts:712`, `packages/build/src/generators/class-schema.ts:753`
The code casts the public `extensionRegistry` option to `MutableExtensionRegistry` and then calls `setSymbolMap()` unconditionally. Existing custom registry implementations that satisfy the published read-only `ExtensionRegistry` interface will now throw at runtime as soon as `configPath` is provided.

4. [P2] Symbol-map seeding resolves registrations by `typeName` only, which is ambiguous across valid multi-extension configs.
`packages/build/src/extensions/symbol-registry.ts:75`, `packages/build/src/extensions/symbol-registry.ts:186`
Each `defineCustomType<T>()` call is mapped back to a runtime registration by scanning for the first matching `typeName`. If two extensions legitimately share the same `typeName` but differ by `tsTypeNames`, this can seed the wrong extension entry into the symbol map.

5. [P2] The new tests do not actually prove the new symbol-based behavior.
Files: `packages/build/src/__tests__/symbol-registry.test.ts:501`, `packages/build/src/__tests__/symbol-registry.test.ts:659`
The `Decimal` integration case can still pass through the legacy name-based resolver because both sides use the same name, and the “symbol beats brand” case uses a single registration whose output is identical whichever path wins. Those tests will not catch regressions in the new symbol path.

PR #276, opened 2026-04-16
Export resolveStaticOptions for config-first schema generation

mike-north:
## Summary

Low-level schema generation functions (`generateSchemasFromDeclaration`, `generateSchemasFromType`, `generateSchemasFromParameter`, `resolveDeclarationMetadata`) now resolve `config.extensions` and `config.metadata` internally via `resolveStaticOptions()`. Previously they only read the deprecated individual fields, forcing callers to pass both `config` and the deprecated fields.

This enables the extensibility SDK to pass just `config: stripeFormSpecConfig` without duplicating the deprecated fields. `resolveStaticOptions` is exported from `@formspec/build/internals` for downstream use.

## Test plan

- [x] 695 build tests pass
- [x] 266 e2e tests pass
- [x] All other packages pass

### DIFF
--- a/.changeset/resolve-static-options.md
+++ b/.changeset/resolve-static-options.md
@@ -1,0 +1,8 @@
+---
+"@formspec/build": patch
+"@formspec/cli": patch
+"@formspec/eslint-plugin": patch
+"formspec": patch
+---
+
+Export resolveStaticOptions and migrate discovered-schema functions to resolve config internally. Low-level functions (generateSchemasFromDeclaration, generateSchemasFromType, etc.) now support the `config` field on StaticSchemaGenerationOptions, eliminating the need to pass deprecated individual fields.
--- a/packages/build/src/__tests__/static-build-context.test.ts
+++ b/packages/build/src/__tests__/static-build-context.test.ts
@@ -388,6 +388,94 @@ describe("static build context", () => {
     });
   });
 
+  // Regression: resolveStaticOptions must read metadata from config.metadata
+  // (not only from the deprecated top-level metadata field).
+  it("generates declaration schemas using metadata policy from config (config-only path)", () => {
+    const context = createStaticBuildContext(entryFixturePath);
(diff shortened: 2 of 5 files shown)


mike-north:
Reviewed PR #276 in an isolated worktree. I found the following issues.

1. [P2] `resolveStaticOptions()` eagerly constructs `config.extensions` before checking the deprecated `extensionRegistry` override.
`packages/build/src/generators/class-schema.ts:736`
This breaks the documented precedence. A caller that passes both can still throw on invalid or duplicate `config.extensions` even though the explicit `extensionRegistry` is supposed to win.

2. [P2] The new config-backed discovered-schema path still never seeds the symbol map from `configPath`.
`packages/build/src/generators/discovered-schema.ts:438`
`generateSchemasFromDeclaration`, `generateSchemasFromType`, `generateSchemasFromParameter`, and `generateSchemasFromReturnType` now resolve `config.extensions`, but they still do not do the `buildSymbolMapFromConfig()` + `setSymbolMap()` follow-through that the high-level schema generation path does. As a result, `defineCustomType<T>()` symbol-based resolution still will not work here for aliased or namespace-imported custom types, so the config-only migration is incomplete.

3. [P3] There is no direct coverage for the newly migrated config-only low-level APIs.
Files: `packages/build/src/__tests__/static-build-context.test.ts:123`, `packages/build/src/__tests__/signature-schema-generation.test.ts:32`
The existing tests still exercise the declaration/type helpers without `config`, so the new bridge and its precedence rules can regress without CI catching it. Please add at least one config-only regression for the declaration/type helpers and one for metadata resolution.

Verification note: I installed the workspace dependencies in the review worktree, but targeted `@formspec/build` verification is currently blocked by existing workspace import/typecheck failures unrelated to this PR (`@formspec/analysis/internal`, `@formspec/core/internals`, plus pre-existing TS errors in `@formspec/build`).

PR #287, opened 2026-04-18
Fix path-targeted constraints on extension-registered custom types

aelliott-stripe:
## Summary

Path-targeted constraint tags (e.g., `@exclusiveMinimum :amount 0`) produced
false `TYPE_MISMATCH` errors when the resolved sub-property was an
extension-registered custom type like `Decimal`. The compiler-backed type
checker in `tsdoc-parser.ts` only understood native TS types and had no
awareness of extension-registered broadenings.

When a path resolves to a registered custom type, the compiler-backed checks
now defer to the downstream IR-based analysis in `semantic-targets.ts`, which
has full broadening and custom constraint awareness. This mirrors the existing
deferral for non-path-targeted broadened types.

## Changes

- Add `isRegisteredCustomTsType()` helper that checks whether a raw `ts.Type`
  is a registered custom type in the extension registry (handles nullable
  unions)
- Restructure `buildCompilerBackedConstraintDiagnostics` so `hasBroadening` is
  computed for both path-targeted and non-path cases
- Add tests for the happy path, nullable variant, and a negative regression
  case

## Test plan

- [x] New unit test: path-targeted `@exclusiveMinimum` on `Decimal` field
  accepts and emits correct schema
- [x] New unit test: same with `Decimal | null` (nullable union stripping)
- [x] New unit test: path-targeted `@exclusiveMinimum` on `string` field still
  rejects with TYPE_MISMATCH
- [x] Existing 715 build tests pass (now 717)
- [x] Existing e2e path-target tests pass

### DIFF
--- a/packages/build/src/__tests__/generate-schemas-config.test.ts
+++ b/packages/build/src/__tests__/generate-schemas-config.test.ts
@@ -184,6 +184,115 @@ describe("generateSchemas with FormSpecConfig", () => {
     }
   });
 
+  describe("path-targeted constraints on broadened custom types", () => {
+    const config: FormSpecConfig = {
+      extensions: [numericExtension],
+      vendorPrefix: "x-formspec",
+    };
+
+    it("allows numeric constraints when path resolves to a broadened custom type", () => {
+      const filePath = writeTempSource(`
+        export type Decimal = string & { __brand: "Decimal" };
+
+        export interface MonetaryAmount {
+          amount: Decimal;
+          currency: string;
+        }
+
+        export interface PaymentForm {
+          /** @exclusiveMinimum :amount 0 */
+          total: MonetaryAmount;
+        }
+      `);
+
+      try {
+        // errorReporting: "throw" guarantees this throws on any TYPE_MISMATCH.
+        const result = generateSchemas({
+          filePath,
+          typeName: "PaymentForm",
+          config,
+          errorReporting: "throw",
+        });
+
+        // The path-targeted constraint should produce an allOf with an
+        // override entry that applies the broadened constraint to `amount`.
+        expect(result.jsonSchema.properties?.["total"]).toMatchObject({
+          allOf: expect.arrayContaining([
+            expect.objectContaining({
+              properties: { amount: { exclusiveMinimum: 0 } },
+            }),
+          ]),
+        });
+      } finally {
+        fs.rmSync(path.dirname(filePath), { recursive: true, force: true });
+      }
+    });
+
+    it("allows numeric constraints when path resolves to a nullable broadened custom type", () => {
+      const filePath = writeTempSource(`
+        export type Decimal = string & { __brand: "Decimal" };
+
+        export interface MonetaryAmount {
+          amount: Decimal | null;
+          currency: string;
+        }
+
+        export interface PaymentForm {
+          /** @exclusiveMinimum :amount 0 */
+          total: MonetaryAmount;
+        }
+      `);
+
+      try {
+        const result = generateSchemas({
+          filePath,
+          typeName: "PaymentForm",
+          config,
+          errorReporting: "throw",
+        });
+
+        expect(result.jsonSchema.properties?.["total"]).toMatchObject({
+          allOf: expect.arrayContaining([
+            expect.objectContaining({
+              properties: { amount: { exclusiveMinimum: 0 } },
+            }),
+          ]),
+        });
+      } finally {
+        fs.rmSync(path.dirname(filePath), { recursive: true, force: true });
+      }
+    });
+
+    it("still rejects numeric constraints when path resolves to a non-numeric, non-custom type", () => {
+      const filePath = writeTempSource(`
+        export type Decimal = string & { __brand: "Decimal" };
+
+        export interface MonetaryAmount {
+          amount: Decimal;
+          currency: string;
+        }
+
+        export interface PaymentForm {
+          /** @exclusiveMinimum :currency 0 */
+          total: MonetaryAmount;
+        }
+      `);
+
+      try {
+        expect(() =>
+          generateSchemas({
+            filePath,
+            typeName: "PaymentForm",
+            config,
+            errorReporting: "throw",
+          })
+        ).toThrow(/TYPE_MISMATCH/);
+      } finally {
+        fs.rmSync(path.dirname(filePath), { recursive: true, force: true });
+      }
+    });
+  });
+
   it("works without config (backward compatibility)", () => {
     const filePath = writeTempSource(`
       export interface SimpleForm {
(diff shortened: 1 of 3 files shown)


packages/build/src/analyzer/tsdoc-parser.ts
```diff
@@ -455,6 +455,51 @@ function hasBuiltinConstraintBroadening(tagName: string, options?: ParseTSDocOpt
...
+  type: ts.Type,
+  options?: ParseTSDocOptions
+): boolean {
+  const registry = options?.extensionRegistry;
+  if (registry === undefined) {
+    return false;
+  }
+
+  const lookup = (candidate: ts.Type): boolean => {
+    const symbol = candidate.getSymbol() ?? candidate.aliasSymbol;
```
mike-north, line 481: Symbol ordering is inverted relative to the codebase convention. `resolveCanonicalSymbol` in `packages/build/src/analyzer/class-analyzer.ts:97` uses `aliasSymbol ?? getSymbol()` with an explicit comment: _"We prefer `aliasSymbol` when present so that aliased types resolve to the declaration site rather than the structural shape."_ The current test fixtures (`type Decimal = string & { __brand: "Decimal" }`) happen to have `getSymbol() === undefined` for the anonymous intersection, so the `??` fallback works today. But for object-shaped aliases like `type Decimal = { value: string; [brand]: "Decimal" }`, `getSymbol()` returns a synthetic `__type` symbol, the alias is never consulted, and `findTypeByName("__type")` fails silently.

```suggestion
    const symbol = candidate.aliasSymbol ?? candidate.getSymbol();
```

packages/build/src/analyzer/tsdoc-parser.ts
```diff
@@ -455,6 +455,51 @@ function hasBuiltinConstraintBroadening(tagName: string, options?: ParseTSDocOpt
...
+    const typeName = symbol?.getName();
+    return typeName !== undefined && registry.findTypeByName(typeName) !== undefined;
```
mike-north, line 483: Only matching by `findTypeByName` gates this fix on the deprecated registration path. The class-analyzer runs three resolvers in sequence ([`class-analyzer.ts:1917-1929`](../blob/92b412a9cfcb6451f1dabe78d257266b28b87c3c/packages/build/src/analyzer/class-analyzer.ts#L1917-L1929)): `resolveRegisteredCustomType` (name) → `resolveSymbolBasedCustomType` (symbol) → `resolveBrandedCustomType` (brand). Users who register custom types via `defineCustomType<T>()` (#275) or the `brand` field (#273) — both added specifically because `tsTypeNames` is unreliable for aliases — will still see the original `TYPE_MISMATCH` false positive on path targets.

At minimum, chain `findTypeBySymbol(symbol)` on the already-resolved `symbol` before falling through. Ideally this deferral would reuse the class-analyzer's resolver chain directly. If a full fix is out of scope, please update the JSDoc's "will not match" caveat to call out brand/symbol registrations and link a follow-up issue.

packages/build/src/analyzer/tsdoc-parser.ts
```diff
@@ -455,6 +455,51 @@ function hasBuiltinConstraintBroadening(tagName: string, options?: ParseTSDocOpt
...
+  };
+
+  if (lookup(type)) {
+    return true;
+  }
+
+  // Handle nullable unions (e.g., Decimal | null)
+  if (type.isUnion()) {
```
mike-north, line 491: This reimplements `stripNullishUnion` from [`packages/analysis/src/ts-binding.ts:23`](../blob/92b412a9cfcb6451f1dabe78d257266b28b87c3c/packages/analysis/src/ts-binding.ts#L23). Using it would also collapse `T | null | undefined` to `T` (which the current `.some(...)` misses), and the flag-check would align with the `(flags & (Null | Undefined)) === 0` idiom used throughout `ts-binding.ts` rather than the double-`!` form.

Minimal fix without a cross-package import:

```suggestion
  // Handle nullable unions (e.g., Decimal | null)
  if (type.isUnion()) {
    return type.types.some(
      (member) =>
        (member.flags & (ts.TypeFlags.Null | ts.TypeFlags.Undefined)) === 0 &&
        lookup(member)
    );
  }
```

Better: export `stripNullishUnion` from `@formspec/analysis/internal` and let it subsume the entire union branch.

packages/build/src/analyzer/tsdoc-parser.ts
```diff
@@ -539,10 +587,15 @@ function buildCompilerBackedConstraintDiagnostics(
       ];
     }
 
+    // If the path resolves to a registered custom type, defer all validation
+    // to the IR-based analysis (semantic-targets.ts) which understands
+    // extension semantics like constraint broadening.
+    hasBroadening = isRegisteredCustomTsType(resolution.type, options);
```
mike-north, line 593: `hasBroadening` here means "the type is extension-registered," not "broadening exists for `tagName` on this type." The sibling non-path branch at line 610 correctly uses `hasBuiltinConstraintBroadening(tagName, options)` which is tag-aware. As written, `@pattern :amount 0` on a numeric `Decimal` field — where no string-pattern broadening is registered — silently passes the compiler-backed check and relies entirely on downstream IR validation, which produces a less specific diagnostic than the capability-aware message this layer would emit.

Consider narrowing the deferral to `findBuiltinConstraintBroadening(customTypeId, tagName)` (mirroring the sibling branch). The accompanying test coverage should include a non-broadenable tag on a custom type to lock this in.

packages/build/src/__tests__/generate-schemas-config.test.ts
```diff
@@ -184,6 +184,115 @@ describe("generateSchemas with FormSpecConfig", () => {
...
+
+      try {
+        expect(() =>
+          generateSchemas({
+            filePath,
+            typeName: "PaymentForm",
+            config,
+            errorReporting: "throw",
+          })
+        ).toThrow(/TYPE_MISMATCH/);
```
mike-north, line 289: A few coverage gaps worth closing while these tests are fresh (they'd also catch the two issues flagged on `tsdoc-parser.ts`):

- **Only `@exclusiveMinimum` is exercised.** A regression that singled out one tag wouldn't be caught. `it.each([["@minimum", ...], ["@maximum", ...], ["@multipleOf", ...]])` would parameterize cheaply.
- **No non-broadenable tag on a custom-type path** (e.g., `@pattern :amount "..."` on `amount: Decimal`). Without this, the compiler-backed layer's over-broad deferral (see the `hasBroadening` comment) slips through.
- **No brand-registered or symbol-registered custom type** (see the `findTypeByName`-only comment). Even if they remain unsupported today, an explicit test asserting the current rejection locks the behavior.
- **Negative assertion is too loose.** `.toThrow(/TYPE_MISMATCH/)` matches any `TYPE_MISMATCH` in the pipeline:

```suggestion
        ).toThrow(/TYPE_MISMATCH.*currency.*exclusiveMinimum|exclusiveMinimum.*currency/);
```

Also consider extracting the near-identical `MonetaryAmount` template into a `monetaryFixture(tag)` helper — the three tests duplicate ~15 lines each.

mike-north (changes requested):

aelliott-stripe:
Handing off to @mike-north

PR #289, opened 2026-04-18
Fix path-targeted broadening for all custom-type registrations

mike-north:
**Alternative to #287.** Fixes the same bug — path-targeted constraints on extension-registered custom types producing spurious `TYPE_MISMATCH` — but covers all three registration mechanisms and consolidates the duplicated custom-type detection. If you merge this, #287 should be closed.

## Why the more holistic fix

The compiler-backed validator in `tsdoc-parser.ts` had no way to recognize extension-registered custom types, so `@exclusiveMinimum :amount 0` where `amount: Decimal` was rejected as a capability mismatch. `class-analyzer.ts` had a rich three-resolver chain (name / symbol / brand) covering the same question for its own IR-building path, but those resolvers were private and the wrong shape for `ts.Type`-only inputs.

PR #287's minimal fix added a new name-only helper to `tsdoc-parser.ts`. That fix:
- Missed brand-registered custom types (per #273), which are the recommended structural detection mechanism.
- Missed symbol-registered custom types (per #275, `defineCustomType<T>()`).
- Used `hasBroadening` as "is a custom type" rather than "has broadening for this tag" — so `@pattern :amount` on a numeric `Decimal` would silently pass the compiler-backed check.
- Inverted the `aliasSymbol ?? getSymbol()` ordering vs. the codebase convention.

This PR addresses all four issues by sharing the detection logic.

## Design

**New**: `packages/build/src/extensions/resolve-custom-type.ts` owns the canonical `ts.Type`-to-registered-custom-type resolution. It tries all three mechanisms in order (name → symbol → brand), strips nullish unions once, and returns the resolved lookup (or `null`). Composite helpers `hasBroadeningForCustomType` and `isTagBroadenedOnTsType` layer the tag-aware broadening check on top.

**Refactor**: `class-analyzer.ts`'s private resolvers (`resolveRegisteredCustomType`, `resolveSymbolBasedCustomType`, `resolveBrandedCustomType`, plus `resolveCanonicalSymbol` and helpers — ~100 lines) are deleted and replaced with a single call to the shared helper.

**Refactor**: `buildCompilerBackedConstraintDiagnostics` collapses its two branches (path-targeted vs. direct-field) into a symmetric `resolve type → check broadening → capability check` flow. Both branches are tag-aware.

**Infrastructure**: `@formspec/analysis/internal` exports `stripNullishUnion` — the single source of truth used across layers.

## Test matrix

Parametrized via `it.each` to cover the grid this class of bug lives in:

| axis | values |
|---|---|
| Registration | `tsTypeNames` (name) · `brand` · `defineCustomType<T>` (symbol) |
| Tag (positive) | `@minimum`, `@maximum`, `@exclusiveMinimum`, `@exclusiveMaximum`, `@multipleOf` |
| Wrapping | bare · `T \| null` · deeply-nested path (`:inner.amount`) |
| Tag (negative) | `@pattern` on numeric `Decimal` (must still reject) |
| Regression | `:currency` path to `string` (must still reject); direct-field capability mismatch (must still reject) |

16 new positive cases + 3 negative regressions. All pass. 742 build tests, 266 e2e, full lint clean.

## Relation to #287

#287 covers the name-based case only and is otherwise correct — the incremental additions here are:
1. Detect brand- and symbol-registered custom types.
2. Make the deferral tag-aware via `findBuiltinConstraintBroadening(customTypeId, tagName)`.
3. Consolidate resolution (class-analyzer's chain + tsdoc-parser's new helper) into one code path.

Closes #287 if merged.

### DIFF
--- a/packages/build/src/__tests__/generate-schemas-config.test.ts
+++ b/packages/build/src/__tests__/generate-schemas-config.test.ts
@@ -3,8 +3,11 @@ import * as fs from "node:fs";
 import * as os from "node:os";
 import { describe, expect, it } from "vitest";
 import type { FormSpecConfig } from "@formspec/config";
-import { generateSchemas } from "../generators/class-schema.js";
+import { defineCustomType, defineExtension } from "@formspec/core/internals";
+import { type ClassSchemas, generateSchemas } from "../generators/class-schema.js";
 import { numericExtension } from "./fixtures/example-numeric-extension.js";
+
+type JsonSchemaObject = Record<string, unknown>;
 
 // ---------------------------------------------------------------------------
 // Helpers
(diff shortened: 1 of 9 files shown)


packages/build/src/extensions/resolve-custom-type.ts
```diff
@@ -0,0 +1,244 @@
...
+import * as ts from "typescript";
+
+import type { ExtensionRegistry, ExtensionTypeLookupResult } from "./registry.js";
+
+/**
+ * Collects all brand identifier texts from an intersection type's computed
+ * property names. Mirrors the helper in `analyzer/class-analyzer.ts`; kept
+ * local so this module has no dependency on the analyzer layer.
+ */
+function collectBrandIdentifiers(type: ts.Type): readonly string[] {
```
mike-north: The PR's consolidation is incomplete. Four helpers are now maintained in two places:

| helper | new copy | existing copy |
|---|---|---|
| `collectBrandIdentifiers` | this file, L11 | `class-analyzer.ts:73` (still used by `isIntegerBrandedType`) |
| `resolveCanonicalSymbol` | this file, L34 | `symbol-registry.ts:164` |
| `extractTypeNodeFromSource` | this file, L63 | `class-analyzer.ts:1728` |
| `getTypeAliasDeclarationFromTypeReference` | this file, L78 | `class-analyzer.ts:3001` |

The "kept local so this module has no dependency on the analyzer layer" rationale is solved by moving these to a neutral module (e.g. `packages/build/src/extensions/ts-type-utils.ts` or `packages/build/src/analyzer/ts-type-utils.ts`) and importing from both sites. The refactor removed the three-resolver chain (~100 lines) — that's a real win — but the changeset claims "consolidate duplicated custom-type resolution" and shipping with four duplicated helpers undermines the claim. Either finish the consolidation as a follow-up commit in this PR or retitle the changeset to "fix path-target broadening + delete the explicit three-resolver chain" so the scope matches.

packages/build/src/__tests__/generate-schemas-config.test.ts
```diff
@@ -184,6 +187,253 @@ describe("generateSchemas with FormSpecConfig", () => {
...
+  // path target (`@exclusiveMinimum :amount 0`) must defer to the IR-layer
+  // validator when the path resolves to an extension-registered custom type
+  // that has broadening for that tag. Previously the compiler-backed
+  // validator rejected these at the capability check, silently shadowing
+  // legitimate broadening.
+  //
+  // This suite exercises the full matrix of registration mechanisms
+  // (name / brand) × tag kinds × wrapping shapes.
+  // =========================================================================
+  describe("path-targeted broadening on registered custom types", () => {
```
mike-north, line 202: The matrix only covers name-based and brand-based registration. Symbol-based registration (via `defineCustomType<T>()` + config-pipeline `setSymbolMap`) is the third mechanism the shared resolver supports — and the one the description emphasizes — but no test exercises that code path through this file. Symbol lookups happen in `symbol-registry.test.ts`, but not in combination with path-targeted broadening.

Recommend adding a third parametrized describe block that registers `Decimal` via a config fixture using `defineCustomType<Decimal>()` and runs the same tag × wrapping grid. If the friction of wiring the symbol map is too high for this PR, at minimum document the gap with an `.skip('not wired: symbol-based path-target broadening')` test carrying a TODO and issue link. Otherwise a regression that only breaks symbol-based detection ships silently.

packages/build/src/__tests__/generate-schemas-config.test.ts
```diff
@@ -184,6 +187,253 @@ describe("generateSchemas with FormSpecConfig", () => {
...
+            export interface MonetaryAmount { amount: Decimal; currency: string; }
+            export interface PaymentForm {
+              /** @${tag} :amount ${argument} */
+              total: MonetaryAmount;
+            }
+          `;
+          // We don't assert the x-prefixed keyword value because the
+          // branded extension uses a different vendor prefix; the test only
+          // needs to confirm generation succeeds without TYPE_MISMATCH.
+          expect(() => runSchema(source, brandBasedConfig)).not.toThrow();
```
mike-north: `expect(() => runSchema(...)).not.toThrow()` passes even if the path override is silently dropped (generation succeeds but no `allOf` entry is emitted). That's a meaningful regression class — a future refactor that breaks brand-based IR emission without breaking deferral would ship green.

Vendor-prefix-agnostic smoke test:

```suggestion
          const result = runSchema(source, brandBasedConfig);
          const total = (result.jsonSchema.properties as JsonSchemaObject | undefined)?.["total"];
          const allOf = (total as { allOf?: readonly JsonSchemaObject[] } | undefined)?.allOf;
          expect(
            allOf?.some((entry) => (entry["properties"] as JsonSchemaObject | undefined)?.["amount"] !== undefined)
          ).toBe(true);
```

Same treatment for the nullable variant at L385.

packages/build/src/__tests__/generate-schemas-config.test.ts
```diff
@@ -184,6 +187,253 @@ describe("generateSchemas with FormSpecConfig", () => {
...
+        // deferral must not kick in and the capability check must reject.
+        const source = `
+          ${nameBasedDecimal}
+          export interface MonetaryAmount { amount: Decimal; currency: string; }
+          export interface PaymentForm {
+            /** @pattern :amount ^[0-9]+$ */
+            total: MonetaryAmount;
+          }
+        `;
+        expect(() => runSchema(source, nameBasedConfig)).toThrow(/TYPE_MISMATCH/);
```
mike-north: The sibling negative test at L401-403 already uses a tighter regex requiring both the segment name and tag name. Apply the same pattern here so the assertion can't pass on an unrelated `TYPE_MISMATCH` from the pipeline:

```suggestion
        expect(() => runSchema(source, nameBasedConfig)).toThrow(
          /TYPE_MISMATCH.*amount.*pattern|pattern.*amount/
        );
```

packages/build/src/__tests__/generate-schemas-config.test.ts
```diff
@@ -3,9 +3,12 @@ import * as fs from "node:fs";
 import * as os from "node:os";
 import { describe, expect, it } from "vitest";
 import type { FormSpecConfig } from "@formspec/config";
-import { generateSchemas } from "../generators/class-schema.js";
+import { defineCustomType, defineExtension } from "@formspec/core/internals";
+import { type ClassSchemas, generateSchemas } from "../generators/class-schema.js";
 import { numericExtension } from "./fixtures/example-numeric-extension.js";
 
+type JsonSchemaObject = Record<string, unknown>;
```
mike-north: `result.jsonSchema` is already typed `JsonSchema2020` ([packages/build/src/json-schema/ir-generator.ts](../blob/00c897fea6def8d31abecaf520b6e2ce3489c16e/packages/build/src/json-schema/ir-generator.ts)) with `properties?: Record<string, JsonSchema2020>` and `allOf?: readonly JsonSchema2020[]`. The local `JsonSchemaObject = Record<string, unknown>` alias plus the `as JsonSchemaObject | undefined` casts scattered through `expectPathOverride` and the nested-path test bypass a usable public type. Importing `JsonSchema2020` would:
- Eliminate the `as JsonSchemaObject` casts on typed property access (e.g., `allOf?.find((entry) => entry.properties?.[segment]?.[keyword] === value)`)
- Retain casts only for vendor-prefixed / unknown-keyword reads
- Reduce the failure mode where a typo in a property name silently falls through the `Record<string, unknown>` index signature

Casts to `Record<string, unknown>` in test code are a sign the production type wasn't threaded through — tighten the seam.

packages/build/src/extensions/resolve-custom-type.ts
```diff
@@ -0,0 +1,244 @@
+import { stripNullishUnion } from "@formspec/analysis/internal";
```
mike-north: Two smaller things:

**Import ordering.** Sibling modules (e.g. [`class-analyzer.ts`](../blob/00c897fea6def8d31abecaf520b6e2ce3489c16e/packages/build/src/analyzer/class-analyzer.ts#L1-L15)) put `import * as ts from "typescript"` first, then `@formspec/*`, then relative. This module reverses the first two.

```suggestion
import * as ts from "typescript";

import { stripNullishUnion } from "@formspec/analysis/internal";

import type { ExtensionRegistry, ExtensionTypeLookupResult } from "./registry.js";
```

**Single-caller composites.** `isTagBroadenedOnTsType` (L229), `hasBroadeningForCustomType` (L211), and `customTypeIdFromLookup` (L198) each have one (or two) caller(s) today. Each adds an export, a doc block, and a level of indirection. Inlining `isTagBroadenedOnTsType` at its `tsdoc-parser.ts:698` call site is ~3 lines; `customTypeIdFromLookup` is a template-literal one-liner. Worth trimming the surface before other callers pick them up.

mike-north (commented):
Self-review addressed in f4f3fb82:

- Consolidation finished: new  hosts the four previously-duplicated helpers; both class-analyzer.ts and resolve-custom-type.ts now import from it.
- Single-caller composites (`isTagBroadenedOnTsType`, `hasBroadeningForCustomType`) removed; broadening check inlined at its single call site in tsdoc-parser.ts.
- Import ordering fixed in resolve-custom-type.ts.
- Symbol-based registration path-target test added (uses `defineCustomType<SymDecimal>()` + configPath pipeline; `typeName` intentionally doesn't match the alias so only the symbol path can satisfy).
- Tests now use the production `JsonSchema2020` type directly instead of the local `JsonSchemaObject` alias.
- Brand-based tests now assert an allOf override entry (vendor-prefix-agnostic) rather than just `.not.toThrow()`.
- @pattern negative test regex tightened to require both segment + tag.
- Deeply-nested path test uses value 42 to avoid incidental default matches.

All 743 build tests + 266 e2e pass. Clean build + lint.

PR #294, opened 2026-04-18
fix(build): bypass synthetic checker for imported integer-branded types

brooks-stripe:
## Summary

- Adds `isIntegerBrandedType` to `analyzer/builtin-brands.ts` — a structural check for the `__integerBrand` symbol in intersection types
- In `tsdoc-parser.ts`, `hasBroadening` now returns `true` for integer-branded types (after stripping nullish unions) with `numeric-comparable` capability, bypassing the synthetic type checker
- In `semantic-targets.ts`, `checkConstraintOnType` now unwraps nullable unions before computing type capabilities at the IR level
- Regression tests for cross-file imports: non-nullable, nullable (`Integer | null`), optional (`score?: Integer`), and negative (`@pattern` must still reject)

## Problem

`@minimum`/`@maximum` on `Integer`/`PositiveInteger` fields sourced from an external module (e.g. `@stripe/extensibility-sdk/stdlib`) produced spurious `TYPE_MISMATCH` errors. Two independent validation layers were affected:

1. **`tsdoc-parser.ts`**: The synthetic TypeScript file used for constraint validation couldn't resolve the imported type name, causing the check to fail even though the type is structurally integer-branded.
2. **`semantic-targets.ts`**: The IR-level `checkConstraintOnType` checked capabilities against the raw union type for nullable fields, failing to recognize the integer member.

This was exposed by stripe/extensibility-sdk#449, which removed `Integer`/`PositiveInteger` from the SDK extension registry, relying on formspec's native `__integerBrand` detection — but that detection only existed in the IR classification path, not the constraint validation paths.

## Changes

| File | What |
|---|---|
| `analyzer/builtin-brands.ts` | New module: `isIntegerBrandedType` extracted here from `extensions/ts-type-utils.ts` — `__integerBrand` is a core concept, not an extension |
| `analyzer/tsdoc-parser.ts` | Import `stripNullishUnion`, wrap `subjectType` before `isIntegerBrandedType` check |
| `analyzer/class-analyzer.ts` | Import `isIntegerBrandedType` from new location |
| `analysis/semantic-targets.ts` | Unwrap nullable unions in `checkConstraintOnType` before capability checks |
| `extensions/ts-type-utils.ts` | Remove `isIntegerBrandedType` (moved to `builtin-brands.ts`) |

## Test plan

- [x] New: "accepts @minimum and @maximum on a doubly-branded integer type imported from another module"
- [x] New: "accepts @minimum on a nullable imported integer type"
- [x] New: "accepts @minimum on an optional imported integer type"
- [x] New: "rejects @pattern on an imported integer type (not string-like)" — negative test pinning bypass scope
- [x] Clarified: locally-declared multi-branded test documents existing behavior, not regression guard
- [x] All 748 build tests pass
- [x] All 312 analysis tests pass
- [x] Typecheck clean

### DIFF
--- a/.changeset/integer-constraint-cross-file.md
+++ b/.changeset/integer-constraint-cross-file.md
@@ -1,6 +1,8 @@
 ---
 "@formspec/build": patch
 "@formspec/analysis": patch
+"@formspec/language-server": patch
+"@formspec/ts-plugin": patch
 ---
 
 Fix TYPE_MISMATCH false positives for numeric constraint tags (`@minimum`, `@maximum`, etc.) on integer-branded types imported from another module, including nullable (`Integer | null`) and optional (`score?: Integer`) variants.
--- a/.changeset/integer-constraint-cross-file.md
+++ b/.changeset/integer-constraint-cross-file.md
@@ -1,9 +1,12 @@
 ---
 "@formspec/build": patch
+"@formspec/analysis": patch
 ---
 
-Fix TYPE_MISMATCH false positives for numeric constraint tags (`@minimum`, `@maximum`, etc.) on integer-branded types imported from another module.
+Fix TYPE_MISMATCH false positives for numeric constraint tags (`@minimum`, `@maximum`, etc.) on integer-branded types imported from another module, including nullable (`Integer | null`) and optional (`score?: Integer`) variants.
 
-Previously, when a type structurally matching `number & { [__integerBrand]: true }` (e.g. `Integer` or `PositiveInteger` from `@stripe/extensibility-sdk/stdlib`) was imported from an external module, the compiler-backed constraint validator rejected it as a capability mismatch because the synthetic TypeScript program used for validation could not resolve the imported type name.
+Two independent validation layers needed fixes:
 
-The private `isIntegerBrandedType` function in `class-analyzer.ts` is now exported from `ts-type-utils.ts` and shared with `tsdoc-parser.ts`, which uses it to bypass the synthetic check for integer-branded types — consistent with how the IR classification path already treats them.
+1. **`tsdoc-parser.ts`** (compiler-backed constraint validation): The synthetic TypeScript program used for validation couldn't resolve imported type names. The `isIntegerBrandedType` bypass now strips nullish unions before checking, so `Integer | null` is handled correctly.
+
+2. **`semantic-targets.ts`** (IR-level constraint validation): `checkConstraintOnType` checked capabilities against the raw `effectiveType`, which is a `union` IR node for nullable fields. Now unwraps nullable unions to the non-null member before computing type capabilities.
--- a/.changeset/integer-constraint-cross-file.md
+++ b/.changeset/integer-constraint-cross-file.md
@@ -1,0 +1,9 @@
+---
+"@formspec/build": patch
+---
+
+Fix TYPE_MISMATCH false positives for numeric constraint tags (`@minimum`, `@maximum`, etc.) on integer-branded types imported from another module.
+
+Previously, when a type structurally matching `number & { [__integerBrand]: true }` (e.g. `Integer` or `PositiveInteger` from `@stripe/extensibility-sdk/stdlib`) was imported from an external module, the compiler-backed constraint validator rejected it as a capability mismatch because the synthetic TypeScript program used for validation could not resolve the imported type name.
+
+The private `isIntegerBrandedType` function in `class-analyzer.ts` is now exported from `ts-type-utils.ts` and shared with `tsdoc-parser.ts`, which uses it to bypass the synthetic check for integer-branded types — consistent with how the IR classification path already treats them.
--- a/packages/analysis/src/semantic-targets.ts
+++ b/packages/analysis/src/semantic-targets.ts
@@ -1144,15 +1144,27 @@ function checkConstraintOnType(
   extensionRegistry: ConstraintRegistryLike | undefined
 ): void {
   const effectiveType = dereferenceAnalysisType(type, typeRegistry);
+  // For nullable unions (e.g. Integer | null), unwrap to the non-null member
+  // so that constraint compatibility checks work the same as for the non-null
+  // variant. This mirrors the stripNullishUnion pattern used in ts-binding.ts.
+  const unwrapped =
+    effectiveType.kind === "union"
+      ? (() => {
+          const nonNull = effectiveType.members
+            .map((m) => dereferenceAnalysisType(m, typeRegistry))
+            .filter((m) => !isNullType(m));
+          return nonNull.length === 1 && nonNull[0] !== undefined ? nonNull[0] : effectiveType;
+        })()
+      : effectiveType;
   const isNumber =
-    effectiveType.kind === "primitive" &&
-    ["number", "integer", "bigint"].includes(effectiveType.primitiveKind);
-  const isString = effectiveType.kind === "primitive" && effectiveType.primitiveKind === "string";
-  const isArray = effectiveType.kind === "array";
-  const isEnum = effectiveType.kind === "enum";
+    unwrapped.kind === "primitive" &&
+    ["number", "integer", "bigint"].includes(unwrapped.primitiveKind);
+  const isString = unwrapped.kind === "primitive" && unwrapped.primitiveKind === "string";
+  const isArray = unwrapped.kind === "array";
+  const isEnum = unwrapped.kind === "enum";
   const arrayItemType =
-    effectiveType.kind === "array"
-      ? dereferenceAnalysisType(effectiveType.items, typeRegistry)
+    unwrapped.kind === "array"
+      ? dereferenceAnalysisType(unwrapped.items, typeRegistry)
       : undefined;
   const isStringArray =
     arrayItemType?.kind === "primitive" && arrayItemType.primitiveKind === "string";
--- a/packages/build/src/__tests__/integer-type.test.ts
+++ b/packages/build/src/__tests__/integer-type.test.ts
@@ -41,12 +41,18 @@ function generateSchemasOrThrow(options: Omit<GenerateSchemasOptions, "errorRepo
  * `afterAll`.
  */
 let fixturePath: string;
+let importingFixturePath: string;
 let tmpDir: string;
 
 const FIXTURE_SOURCE = [
   // Local unique-symbol brand — structurally identical to @formspec/core's export.
   "declare const __integerBrand: unique symbol;",
   "type Integer = number & { readonly [__integerBrand]: true };",
+  "",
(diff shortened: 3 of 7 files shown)


mike-north (commented):
Thanks Brooks — real bug, well-characterized, and the regression test genuinely pins it down. I ran the cross-file test with `tsdoc-parser.ts` reverted and confirmed it fails with exactly the claimed `TYPE_MISMATCH` diagnostic; the fix resolves it. Lint + typecheck are clean on changed files. Below is consolidated feedback from TS-expert, simplicity, testing, code-quality, and architecture passes — none are blockers, but a few are worth addressing before merge.

## Concerns

### 1. Union types still regress — `Integer | null`, `Integer | undefined`

`isIntegerBrandedType` checks `type.isIntersection()`, so a union like `Integer | null` (a very common shape for optional SDK fields) returns `false` and the spurious `TYPE_MISMATCH` resurfaces. The IR-layer path had the same limitation, but moving the helper to a shared utility is the right moment to consider it. Suggest: if `type.isUnion()`, check that every non-nullish member is integer-branded.

A regression test with `year: MultiBrandedInteger | null` and `@minimum`/`@maximum` would lock this down.

### 2. Capability gate is too narrow — `@multipleOf` will still break

The bypass gate is `definition.capabilities.includes(\"numeric-comparable\")`, which covers `@minimum` / `@maximum` / `@exclusiveMinimum` / `@exclusiveMaximum`. But `@multipleOf` uses a different capability (`numeric-divisible` or similar) and would still hit the synthetic-checker failure on imported integer types. Either broaden to \"any numeric capability applicable to `number`\" or add an explicit allow-list. Worth aligning with whatever `hasBuiltinConstraintBroadening` does for locally-declared integers.

### 3. Two parallel broadening mechanisms, no shared abstraction

The `target === null` branch in `tsdoc-parser.ts` now has two independent paths answering the same question (\"is `tagName` broadened onto this type?\"): the new `isIntegerBrandedType + capability` check, and `hasBuiltinConstraintBroadening` (extension-registry lookup). The comment block above calls this the \"unified broadening check\" — the new code breaks that unification. A third built-in brand (string-branded IDs, duration, etc.) would add a third ad-hoc branch.

The natural refactor: a single `resolveKnownCapabilities(type, checker): Capability[] | null` consulted before synthesis, with built-in brands registering into the same capability table as extension types. Not required for this PR, but the next brand is the inflection point — worth noting as a follow-up.

### 4. File placement — `extensions/` is the wrong module

`isIntegerBrandedType` landed in `packages/build/src/extensions/ts-type-utils.ts`, but `extensions/` is documented as the home for the extension registry / user-registered custom types — and the JSDoc on the old helper even says so (\"Extension-registered custom types... are resolved separately\"). `__integerBrand` is a formspec core concept, not an extension.

Suggest relocating to something like `analyzer/builtin-brands.ts` (or keep it in `class-analyzer.ts` and re-export). Ideally co-located with the brand → capability mapping, since detection and capability policy change together and have no other consumers. `collectBrandIdentifiers` can stay in `ts-type-utils.ts` — it's a generic TS helper.

### 5. Locally-declared test doesn't guard the fix

I verified by reverting only `tsdoc-parser.ts`: the new `MultiBrandedConstrainedConfig` (local declaration) test **passes even without the fix** — the class-analyzer path already handled double-branded local types. Only the `CrossFileConstrainedConfig` test is a real regression guard. Not wrong to include both, but either add a comment clarifying that the local case documents existing behavior, or drop it to avoid implying both guard the regression.

### 6. Missing negative test — pin the scope of the bypass

No test confirms that a non-numeric tag (e.g. `@pattern`) on an imported integer-branded type **still produces** a `TYPE_MISMATCH`. Adding one would lock in the bypass's scope to numeric-comparable tags and prevent accidental future over-broadening. Combined with a `@multipleOf` case, this would close the capability-gate concern from #2.

## Nits

- `ts-type-utils.ts:13-14` — single-line `if (...) return ...;` without braces differs from the braced form of the original helper in `class-analyzer.ts`. Minor, matches local file style.
- Consider `@internal` / release tag on the exported `isIntegerBrandedType` so API Extractor doesn't treat it as public surface.
- `expect(yearSchema).toMatchObject({ minimum, maximum })` permits extra properties — fine here since the regression was a throw, but exact matching is slightly stronger.

## Strengths

- Cross-module fixture (`importing-fixture.ts` + `types.ts`) precisely reproduces the real-world SDK failure; great test scoping.
- Multi-brand fixture (`__integerBrand` + `__stripeType`) guards against a naive single-symbol check regressing.
- Deduplicating the helper instead of copy-pasting is the right call.
- Changeset is clear about user-visible behavior and the mechanism.
- Capability gate on the bypass prevents accidental over-broadening of non-numeric tags on integers.

## Verification

- `pnpm --filter @formspec/build run test -- integer-type.test.ts` → all pass on PR branch.
- With `tsdoc-parser.ts` changes reverted: `CrossFileConstrainedConfig` test fails with `TYPE_MISMATCH` (confirms regression guard). `MultiBrandedConstrainedConfig` (local) still passes (see #5).
- Lint + typecheck on changed files: clean.

Happy to land as-is if you want to ship the SDK unblock fast and file follow-ups for #1–#4, but #1 (union gap) and #5/#6 (test scope) would be worth handling in this PR if quick.

jar-stripe:
## Edge case: nullable/optional imported integer types

The current fix handles non-nullable imported integers, but nullable (`Integer | null`) and optional (`score?: Integer`) imported integer fields still produce spurious `TYPE_MISMATCH` errors. There are two independent validation layers that both need to handle this:

**1. `tsdoc-parser.ts`** — The `isIntegerBrandedType` bypass doesn't fire because `subjectType` is a union (`Integer | null`), not an intersection. Fix: strip nullish before the check.

**2. `semantic-targets.ts`** — The IR-level `checkConstraintOnType` checks `isNumber` against `effectiveType`, which is a `union` IR node for nullable fields. Fix: unwrap nullable unions to the non-null member before computing `isNumber`/`isString`/etc.

Here's the full diff (all tests pass including 2 new test cases):

```diff
diff --git a/packages/analysis/src/semantic-targets.ts b/packages/analysis/src/semantic-targets.ts
index c652353..921f761 100644
--- a/packages/analysis/src/semantic-targets.ts
+++ b/packages/analysis/src/semantic-targets.ts
@@ -1144,15 +1144,27 @@ function checkConstraintOnType(
   extensionRegistry: ConstraintRegistryLike | undefined
 ): void {
   const effectiveType = dereferenceAnalysisType(type, typeRegistry);
+  // For nullable unions (e.g. Integer | null), unwrap to the non-null member
+  // so that constraint compatibility checks work the same as for the non-null
+  // variant. This mirrors the stripNullishUnion pattern used in ts-binding.ts.
+  const unwrapped =
+    effectiveType.kind === "union"
+      ? (() => {
+          const nonNull = effectiveType.members
+            .map((m) => dereferenceAnalysisType(m, typeRegistry))
+            .filter((m) => !isNullType(m));
+          return nonNull.length === 1 && nonNull[0] !== undefined ? nonNull[0] : effectiveType;
+        })()
+      : effectiveType;
   const isNumber =
-    effectiveType.kind === "primitive" &&
-    ["number", "integer", "bigint"].includes(effectiveType.primitiveKind);
-  const isString = effectiveType.kind === "primitive" && effectiveType.primitiveKind === "string";
-  const isArray = effectiveType.kind === "array";
-  const isEnum = effectiveType.kind === "enum";
+    unwrapped.kind === "primitive" &&
+    ["number", "integer", "bigint"].includes(unwrapped.primitiveKind);
+  const isString = unwrapped.kind === "primitive" && unwrapped.primitiveKind === "string";
+  const isArray = unwrapped.kind === "array";
+  const isEnum = unwrapped.kind === "enum";
   const arrayItemType =
-    effectiveType.kind === "array"
-      ? dereferenceAnalysisType(effectiveType.items, typeRegistry)
+    unwrapped.kind === "array"
+      ? dereferenceAnalysisType(unwrapped.items, typeRegistry)
       : undefined;
   const isStringArray =
     arrayItemType?.kind === "primitive" && arrayItemType.primitiveKind === "string";
diff --git a/packages/build/src/__tests__/integer-type.test.ts b/packages/build/src/__tests__/integer-type.test.ts
index b945fdc..974b35e 100644
--- a/packages/build/src/__tests__/integer-type.test.ts
+++ b/packages/build/src/__tests__/integer-type.test.ts
@@ -121,6 +121,16 @@ const IMPORTING_FIXTURE_SOURCE = [
   "  /** @minimum 2000 @maximum 2026 */",
   "  year: MultiBrandedInteger;",
   "}",
+  "",
+  "export interface CrossFileNullableConfig {",
+  "  /** @minimum 0 */",
+  "  score: MultiBrandedInteger | null;",
+  "}",
+  "",
+  "export interface CrossFileOptionalConfig {",
+  "  /** @minimum 0 */",
+  "  score?: MultiBrandedInteger;",
+  "}",
 ].join("\n");
 
 beforeAll(() => {
@@ -442,5 +452,33 @@ describe("builtin Integer type", () => {
         maximum: 2026,
       });
     });
+
+    it("accepts @minimum on a nullable imported integer type", () => {
+      const result = generateSchemasOrThrow({
+        filePath: importingFixturePath,
+        typeName: "CrossFileNullableConfig",
+      });
+
+      const properties = result.jsonSchema.properties as Record<string, unknown>;
+      const scoreSchema = properties["score"] as Record<string, unknown>;
+
+      expect(scoreSchema).toMatchObject({
+        minimum: 0,
+      });
+    });
+
+    it("accepts @minimum on an optional imported integer type", () => {
+      const result = generateSchemasOrThrow({
+        filePath: importingFixturePath,
+        typeName: "CrossFileOptionalConfig",
+      });
+
+      const properties = result.jsonSchema.properties as Record<string, unknown>;
+      const scoreSchema = properties["score"] as Record<string, unknown>;
+
+      expect(scoreSchema).toMatchObject({
+        minimum: 0,
+      });
+    });
   });
 });
diff --git a/packages/build/src/analyzer/tsdoc-parser.ts b/packages/build/src/analyzer/tsdoc-parser.ts
index 870f8e6..59e3dd2 100644
--- a/packages/build/src/analyzer/tsdoc-parser.ts
+++ b/packages/build/src/analyzer/tsdoc-parser.ts
@@ -45,6 +45,7 @@ import {
   getTagDefinition,
   hasTypeSemanticCapability,
   normalizeFormSpecTagName,
+  stripNullishUnion,
   parseConstraintTagValue,
   parseDefaultValueTagValue,
   parseTagSyntax,
@@ -700,7 +701,7 @@ function buildCompilerBackedConstraintDiagnostics(
     if (target === null) {
       if (
         subjectType !== undefined &&
-        isIntegerBrandedType(subjectType) &&
+        isIntegerBrandedType(stripNullishUnion(subjectType)) &&
         definition.capabilities.includes("numeric-comparable")
       ) {
         return true;
```

PR #297, opened 2026-04-18
fix(build): prevent imported type from poisoning sibling field constraints

brooks-stripe:
## Problem

When an interface (or type alias) contains both an imported type field and a plain type field with constraints, the plain field produces spurious `TYPE_MISMATCH` errors:

```typescript
import { Integer } from '@stripe/extensibility-sdk/stdlib';

export interface VehicleFields {
  /** @minimum 2000 */
  year: Integer;

  /** @minLength 17 @maxLength 17 */
  vin: string;  // ❌ TYPE_MISMATCH: Tag "@minLength" received an invalid argument for integer.
}
```

`buildSupportingDeclarations` excludes the entire interface from the synthetic checker's program because it references an imported name (`Integer`). Without the interface declaration, the synthetic checker has no type context for `vin`, so `@minLength`/`@maxLength` fail.

This affects any interface or type alias mixing imported types with constrained non-imported types.

## Fix

For interface declarations and type alias declarations with type-literal bodies that reference imported names, rewrite the imported member type annotations to `unknown` instead of dropping the entire declaration. The synthetic checker can then still resolve non-imported member types correctly.

```
// Before (excluded entirely):
// export interface VehicleFields { year: Integer; vin: string; }

// After (rewritten):
export interface VehicleFields { year: unknown; vin: string; }
```

Other statement types that reference imports (classes, variables, non-type-literal type aliases) are still excluded as before.

Also deduplicates `statementReferencesImportedName` and `typeAnnotationReferencesImport` into a single `astReferencesImportedName` helper to eliminate the copy-pasted AST traversal logic.

## Test plan

- [x] New: "does not poison sibling string fields with imported integer type constraints" — `CrossFileMixedConfig` with `year: MultiBrandedInteger` + `vin: string`
- [x] New: "does not poison sibling string fields in type alias declarations" — `CrossFileMixedTypeAlias` with the same shape as a type alias
- [x] All 753 build tests pass (including previously-passing cross-module tests in `nounchecked-index-access.test.ts`)
- [x] Typecheck clean

### DIFF
--- a/.changeset/fix-sibling-field-poisoning.md
+++ b/.changeset/fix-sibling-field-poisoning.md
@@ -1,0 +1,7 @@
+---
+"@formspec/build": patch
+---
+
+Fix constraint validation on sibling fields when an interface contains an imported type. Previously, an interface like `{ year: Integer; vin: string }` where `Integer` was imported from another module would be entirely excluded from the synthetic checker's supporting declarations. This caused spurious TYPE_MISMATCH errors on string constraint tags (`@minLength`, `@maxLength`) applied to non-imported sibling fields.
+
+The synthetic program now rewrites imported member types to `unknown` instead of dropping the entire interface, preserving type context for non-imported siblings.
--- a/.changeset/fix-sibling-field-poisoning.md
+++ b/.changeset/fix-sibling-field-poisoning.md
@@ -1,8 +1,5 @@
 ---
 "@formspec/build": patch
-"@formspec/cli": patch
-"@formspec/eslint-plugin": patch
-"formspec": patch
 ---
 
 Fix constraint validation on sibling fields when an interface contains an imported type. Previously, an interface like `{ year: Integer; vin: string }` where `Integer` was imported from another module would be entirely excluded from the synthetic checker's supporting declarations. This caused spurious TYPE_MISMATCH errors on string constraint tags (`@minLength`, `@maxLength`) applied to non-imported sibling fields.
--- a/.changeset/fix-sibling-field-poisoning.md
+++ b/.changeset/fix-sibling-field-poisoning.md
@@ -1,5 +1,8 @@
 ---
 "@formspec/build": patch
+"@formspec/cli": patch
+"@formspec/eslint-plugin": patch
+"formspec": patch
 ---
 
 Fix constraint validation on sibling fields when an interface contains an imported type. Previously, an interface like `{ year: Integer; vin: string }` where `Integer` was imported from another module would be entirely excluded from the synthetic checker's supporting declarations. This caused spurious TYPE_MISMATCH errors on string constraint tags (`@minLength`, `@maxLength`) applied to non-imported sibling fields.
--- a/packages/build/src/__tests__/integer-type.test.ts
+++ b/packages/build/src/__tests__/integer-type.test.ts
@@ -158,6 +158,30 @@ const IMPORTING_FIXTURE_SOURCE = [
   "   */",
   "  vin: string;",
   "};",
+  "",
+  "// Method signature referencing import — should fall back to excluding",
+  "export interface CrossFileMethodSigConfig {",
+  "  getYear(): MultiBrandedInteger;",
+  "  /** @minLength 1 */",
+  "  vin: string;",
+  "}",
+  "",
+  "// Index signature referencing import — should fall back to excluding",
+  "export interface CrossFileIndexSigConfig {",
+  "  [k: string]: MultiBrandedInteger;",
+  "  /** @minLength 1 */",
+  "  vin: string;",
+  "}",
+  "",
+  "// Generic position — Array<ImportedType>",
+  "export interface CrossFileGenericConfig {",
+  "  items: Array<MultiBrandedInteger>;",
+  "  /** @minLength 1 */",
+  "  label: string;",
+  "}",
+  "",
+  "// Const referencing import — should still be excluded",
+  "export const CROSS_FILE_CONST: MultiBrandedInteger = 0 as unknown as MultiBrandedInteger;",
 ].join("\n");
 
 beforeAll(() => {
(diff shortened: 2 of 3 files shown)


mike-north (commented):
Consolidated review from five specialist passes (code quality, simplicity, tests, TypeScript idioms, architecture). The fix is legitimate and the direction is right — but several issues suggest this PR should not land as-is.

## Blockers / majors

### 1. `ts.isPropertyDeclaration(member)` is an unreachable branch
**Location:** `packages/build/src/analyzer/tsdoc-parser.ts` — `rewriteImportedMemberTypes`, member-filter line.

`getObjectMembers` returns `ts.NodeArray<ts.TypeElement>` (from `InterfaceDeclaration.members` or `TypeLiteralNode.members`). `PropertyDeclaration` is a `ClassElement`, not a `TypeElement` — it can never appear here. The `|| ts.isPropertyDeclaration(member)` disjunct is dead code and misleading about what shapes are handled.

### 2. Non-property `TypeElement`s silently regress
**Same function.** The rewrite handles only `PropertySignature`. Index signatures, method signatures, call/construct signatures, and accessors are left untouched — but the outer `astReferencesImportedName` check already matched the whole statement, so control flow falls through to the rewrite branch. Net effect: an interface like
```ts
interface X { [k: string]: Integer; foo: string }
// or
interface X { getYear(): Integer; /** @minLength 1 */ vin: string }
```
previously dropped the whole declaration; now it emits text containing `Integer` into the synthetic file with no import, producing `Cannot find name 'Integer'` and breaking the synthetic compile. This is a **regression** vs. pre-PR behavior for these shapes.

**Fix options:**
- (a) Broaden the member-type extraction to `PropertySignature | IndexSignatureDeclaration | MethodSignature | GetAccessorDeclaration | SetAccessorDeclaration`, and for `MethodSignature` also consider parameter types.
- (b) When the member kind isn't rewritable but references an import, fall back to the old "exclude the whole declaration" behavior rather than emitting invalid text.

### 3. Test coverage is narrow relative to the rewrite's blast radius
The two new tests (`CrossFileMixedConfig`, `CrossFileMixedTypeAlias`) pin the happy path and sibling assertions well, and the regression-first commit history (`d5366e4c`) is good discipline. Gaps:
- **No negative test** that a const / var / function / class referencing an imported name is still excluded (the PR claims this; nothing verifies it).
- **No class test** — same failure mode, same-shape remedy, explicitly out of scope by the PR's own description. Worth pinning current behavior with a test so the next person finds a breadcrumb.
- **Nested type literals**: `interface X { nested: { bar: Integer; baz: string } }` — does the rewrite descend?
- **Generic position**: `interface X { items: Array<Integer>; count: number }`.
- **`readonly` / non-null unions**: only the `| null` case is covered (via `CrossFileNullableConfig`).
- **Method signatures**, **index signatures**.
- **Non-type-literal alias bodies**: `type Foo = Pick<Bar,'x'> & { vin: string }` — alias is dropped wholesale today; confirm and test.

### 4. Text-slice rewriting should be a TS transformer
**Location:** `rewriteImportedMemberTypes` — the `replacements.reverse()` + `result.slice(...) + "unknown" + result.slice(...)` block.

Manual offset arithmetic on `getText()` output is unusual in a codebase that already depends on `typescript`. `ts.transform` + `ts.factory.createKeywordTypeNode(ts.SyntaxKind.UnknownKeyword)` + `ts.createPrinter().printNode(...)` is ~15 lines, handles all future TypeElement kinds automatically (fixing #2 incidentally), preserves `readonly` / `?` / trivia, and removes the "reverse-order to preserve offsets" comment. The current approach is also vulnerable to a future switch from `getText(sourceFile)` to `getFullText()` silently breaking offsets.

## Architectural (address in follow-up, not blocking)

### 5. Second carve-out in the synthetic-checker pattern
This is the second declaration-level patch (after #294's integer-branded bypass) on top of a design whose root limitation — inability to resolve cross-module references — is already solved elsewhere. `ProgramContext` wires a real host `ts.Program` / `ts.TypeChecker` through `program.ts`, and PR #300 is extending that access to extensions via `resolvePayload(type, checker)`. Each new "imported type reference" surface (classes, intersections, method params, mapped types…) will require another rewrite rule.

**Recommendation:** Open a follow-up issue to replace `checkSyntheticTagApplication` with a host-checker-driven capability check. The direct-field path already has `subjectType: ts.Type` in `options` — the synthetic textual check is redundant for those cases. Keep the synthetic program (if needed at all) only for evaluating constructed expressions, and model imported types as ambient `unknown` aliases *once* at the top of the synthetic file rather than rewriting statement-by-statement. Worth flagging before a third carve-out lands.

### 6. `buildSupportingDeclarations` now mixes three responsibilities
Filter imports, pass through clean declarations, and partially rewrite interface/type-alias declarations by text-splicing. Consider splitting into `selectSupportingStatements(...)` (returns `{keep, rewrite}` partition) and `rewriteStatementToText(...)` — isolates AST-shape knowledge and makes the "add classes next" extension point explicit.

## Minors / nits

- `replacements.reverse()` mutates the local array in place. Harmless today; prefer `[...replacements].reverse()` or a descending-index loop.
- `getObjectMembers` is used exactly once; inlining it removes a helper with no readability loss unless a second caller is planned.
- `astReferencesImportedName` rename is a clean dedup win. `subtreeReferencesImportedName` would read more precisely at call sites, but this is a nit.
- A short comment in `rewriteImportedMemberTypes` tying `statement.getText(sourceFile)` to `statement.getStart(sourceFile)` would protect the offset invariant from a future `getFullText()` swap.
- Changeset is accurate; the `astReferencesImportedName` consolidation is worth landing on its own regardless.

## Strengths

- Bug is real, well-characterized, and reproduced by a pre-fix failing test.
- Rewriting to `unknown` is the right structural placeholder — can't accidentally satisfy any constraint capability.
- Dedup of the two AST traversals is a genuine simplification.
- Happy-path assertions are structural (`toMatchObject({minLength: 17, maxLength: 17})`), not snapshot/tautological.

## Recommendation

**Request changes.** Resolve items 1 and 2 (dead branch + regression for non-property TypeElements) before merge. Prefer the `ts.transform` rewrite (item 4) since it naturally subsumes item 2. Add at minimum a class test, a const/var "still excluded" test, and a method-signature test. Open a follow-up for the host-program consolidation (item 5) — that's the strategic direction this and #294 both point at.

mike-north (approved):

PR #300, opened 2026-04-19
feat(core): add extractPayload callback to CustomTypeRegistration

brooks-stripe:
## Problem

Extensions that register custom types (via `defineCustomType`) have no way to carry type-level information from TypeScript analysis into JSON Schema generation. The `payload` on custom type nodes is always `null`.

This blocks generic types like `Ref<T>` where the generic argument `T` determines the JSON Schema output (e.g., a referenced type tag). Without the payload, `toJsonSchema` can't produce the correct structured schema.

## Fix

Add an optional `extractPayload` callback to `CustomTypeRegistration`:

```typescript
readonly extractPayload?: (
  type: unknown,   // ts.Type
  checker: unknown // ts.TypeChecker
) => ExtensionPayloadValue;
```

When defined, the class analyzer calls `extractPayload(type, checker)` after matching a custom type and stores the result as the node's `payload`. This payload then flows through to `toJsonSchema(payload, vendorPrefix)` during schema generation. Errors thrown by the callback are attributed to the extension (type name + extension ID).

### Why a callback with `unknown` parameters?

The information needed to extract a payload flows through three layers:

1. **`@formspec/core`** defines `CustomTypeRegistration`. This package has no TypeScript compiler dependency — it can't reference `ts.Type` or `ts.TypeChecker`.
2. **`@formspec/build`** runs the class analyzer and has access to the resolved `ts.Type` for each field. It calls `extractPayload` and passes the TypeScript type and checker.
3. **The extension** (the consumer defining the custom type) knows what specific information to extract. For example, a `Ref<T>` extension knows to look up the `type` property's string literal value from the generic argument.

Without a callback, the class analyzer would need to know what to extract for every possible custom type — which it can't, since custom types are defined by extensions. The `unknown` parameters are the cost of the interface living in `@formspec/core` (lightweight, no TS compiler dependency) while the actual values come from `@formspec/build` (which depends on the TS compiler). Extension authors cast back to `ts.Type` and `ts.TypeChecker` — they know what they're receiving because the contract is documented.

### Contract

The TSDoc on `extractPayload` documents:
- **Purity** — must be a pure function of `(type, checker)`, no I/O or shared state
- **Multiplicity** — may be invoked multiple times for the same type (no memoization)
- **JSON-serializability** — must return `ExtensionPayloadValue`, not live compiler objects
- **Error semantics** — errors are attributed to the extension and reported as build errors
- **`undefined` vs `null`** — `undefined` return is coerced to `null`
- **Host program** — both parameters originate from the host program's type checker

## Test plan

- [x] New: `extractPayload` passes extracted payload through to `toJsonSchema`
- [x] New: `null` payload when `extractPayload` is not defined (backwards compatible)
- [x] New: `extractPayload` receives the union type for optional fields
- [x] New: errors from `extractPayload` are attributed to the extension
- [x] New: `undefined` return is coerced to `null`
- [x] All 764 build tests pass
- [x] Typecheck clean

### DIFF
--- a/.changeset/resolve-payload-custom-types.md
+++ b/.changeset/resolve-payload-custom-types.md
@@ -1,0 +1,9 @@
+---
+"@formspec/core": minor
+"@formspec/build": minor
+"@formspec/cli": minor
+"@formspec/eslint-plugin": minor
+"formspec": minor
+---
+
+Add optional `resolvePayload` callback to `CustomTypeRegistration`. When defined, the callback is invoked during type analysis with the TypeScript type and checker, and its return value is stored as the custom type node's `payload`. This payload is then passed to `toJsonSchema` during schema generation, enabling extensions to carry type-level information (e.g., a generic argument's resolved literal value) through to JSON Schema output.
--- a/.changeset/resolve-payload-custom-types.md
+++ b/.changeset/resolve-payload-custom-types.md
@@ -12,4 +12,4 @@
 "formspec": minor
 ---
 
-Add optional `resolvePayload` callback to `CustomTypeRegistration`. When defined, the callback is invoked during type analysis with the TypeScript type and checker, and its return value is stored as the custom type node's `payload`. This payload is then passed to `toJsonSchema` during schema generation, enabling extensions to carry type-level information (e.g., a generic argument's resolved literal value) through to JSON Schema output.
+Add optional `extractPayload` callback to `CustomTypeRegistration`. When defined, the callback is invoked during type analysis with the TypeScript type and checker, and its return value is stored as the custom type node's `payload`. This payload is then passed to `toJsonSchema` during schema generation, enabling extensions to carry type-level information (e.g., a generic argument's resolved literal value) through to JSON Schema output.
--- a/.changeset/resolve-payload-custom-types.md
+++ b/.changeset/resolve-payload-custom-types.md
@@ -3,6 +3,12 @@
 "@formspec/build": minor
 "@formspec/cli": minor
 "@formspec/eslint-plugin": minor
+"@formspec/analysis": minor
+"@formspec/config": minor
+"@formspec/dsl": minor
+"@formspec/language-server": minor
+"@formspec/runtime": minor
+"@formspec/ts-plugin": minor
 "formspec": minor
 ---
 
--- a/packages/build/src/__tests__/resolve-payload.test.ts
+++ b/packages/build/src/__tests__/resolve-payload.test.ts
@@ -1,0 +1,212 @@
+/**
+ * Tests for the `resolvePayload` callback on custom type registrations.
+ *
+ * Verifies that:
+ *   1. `resolvePayload` is called during type analysis and its return value
+ *      flows through to `toJsonSchema` as the `payload` argument.
+ *   2. The callback receives the TypeScript type and checker, enabling
+ *      extraction of type-level information (e.g., generic argument literals).
+ *   3. Optional (nullable) fields pass the union type correctly.
+ */
+
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
+import { defineCustomType, defineExtension } from "@formspec/core/internals";
+import { generateSchemas, type GenerateSchemasOptions } from "../generators/class-schema.js";
+import { createExtensionRegistry } from "../extensions/index.js";
+
+function generateSchemasOrThrow(options: Omit<GenerateSchemasOptions, "errorReporting">) {
+  return generateSchemas({
+    ...options,
+    errorReporting: "throw",
+  });
+}
+
+let tmpDir: string;
+
+beforeAll(() => {
+  tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "formspec-resolve-payload-"));
+});
+
+afterAll(() => {
+  if (tmpDir) {
+    fs.rmSync(tmpDir, { recursive: true, force: true });
+  }
+});
+
+function writeFixture(name: string, lines: string[]): string {
+  const filePath = path.join(tmpDir, name);
+  fs.writeFileSync(filePath, lines.join("\n"));
+  return filePath;
+}
+
+describe("resolvePayload", () => {
+  it("passes resolved payload through to toJsonSchema", () => {
+    const toJsonSchema = vi.fn((_payload, _vendorPrefix) => ({
+      type: "string",
+      "x-ref-target": _payload,
+    }));
+
+    const registry = createExtensionRegistry([
+      defineExtension({
+        extensionId: "x-test/ref",
+        types: [
+          defineCustomType({
+            typeName: "TestRef",
+            tsTypeNames: ["TestRef"],
+            resolvePayload: (type: unknown, checker: unknown) => {
+              const tsType = type as { getProperty(name: string): unknown | undefined };
+              const tsChecker = checker as {
+                getTypeOfSymbol(symbol: unknown): {
+                  isStringLiteral(): boolean;
+                  value: string;
+                };
+              };
+              const prop = tsType.getProperty("target");
+              if (!prop) return null;
+              const propType = tsChecker.getTypeOfSymbol(prop);
+              return propType.isStringLiteral() ? propType.value : null;
+            },
+            toJsonSchema,
+          }),
+        ],
+      }),
+    ]);
+
+    const filePath = writeFixture("resolve-payload-basic.ts", [
+      "type TestRef<T extends string> = {",
+      "  id: string;",
+      "  target: T;",
+      "};",
+      "",
+      "export interface Config {",
+      '  ref: TestRef<"customer">;',
+      "}",
+    ]);
+
+    const { jsonSchema } = generateSchemasOrThrow({
+      filePath,
+      typeName: "Config",
+      extensionRegistry: registry,
+    });
+
+    expect(toJsonSchema).toHaveBeenCalledWith("customer", expect.any(String));
+
+    const props = jsonSchema.properties ?? {};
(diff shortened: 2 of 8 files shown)


mike-north (commented):
Consolidated review from five specialist passes (code quality, API design, tests, TypeScript idioms, architecture). The change solves a real blocker (generic wrapper types like `Ref<T>` producing empty schemas) and is backwards-compatible. The core concerns are around contract tightness and a quiet architectural shift in `@formspec/core`'s boundary.

## Majors — worth addressing before merge

### 1. Error attribution at the call site
**Location:** `packages/build/src/analyzer/class-analyzer.ts:1710`.

`customTypeLookup.registration.resolvePayload?.(type, checker) ?? null` is called bare. An extension-authored callback that throws (NPE on `valueDeclaration.foo`, misused API, assumption violated) will crash schema generation with a stack trace pointing into `resolveTypeNode` and no mention of which extension / type name is at fault. This matches the existing no-catch pattern on `toJsonSchema`, but both sites deserve attribution. Either wrap here and propagate a contextual error (`"resolvePayload for custom type \"${typeName}\" in extension \"${extensionId}\" threw: ..."`, with `cause`), or file a follow-up to wrap both sites uniformly.

### 2. Naming: `resolvePayload` overloads an already-loaded verb
**Location:** `packages/core/src/extensions/index.ts`.

"Resolve" in this codebase already means *runtime async data-source fetching* (`defineResolvers`, `ResolverRegistry`, dynamic-enum resolvers). Reusing the verb for a *synchronous build-time TS type inspection* callback muddies the vocabulary. Sibling hooks use action verbs (`toJsonSchema`, `isApplicableToType`). `extractPayload` reads precisely — the callback *extracts* information from the TS type — and `computePayload` is also fine. Low-cost to change now (`@public` surface, but no in-tree consumers yet).

### 3. `unknown`/`unknown` parameters leak type-safety through the public API
**Location:** `packages/core/src/extensions/index.ts` (signature of `resolvePayload`).

Every consumer will write `const tsType = type as ts.Type; const tsChecker = checker as ts.TypeChecker;`. This preserves `@formspec/core`'s zero-runtime-dep on `typescript`, but at the cost of:
- No compile-time anchor for "which TS version does this contract assume?"
- No signal that the two params must come from the *same* program (see #6 below)
- Users can pass any `unknown`, including values from a different `ts` module instance

**Recommendation:** introduce symbol-branded opaque handles in core, matching the project's own branded-type convention (`declare const __tsTypeBrand: unique symbol`):

```ts
declare const __tsTypeBrand: unique symbol;
declare const __tsCheckerBrand: unique symbol;
export type TsTypeHandle = unknown & { readonly [__tsTypeBrand]: "ts.Type" };
export type TsCheckerHandle = unknown & { readonly [__tsCheckerBrand]: "ts.TypeChecker" };

readonly resolvePayload?: (type: TsTypeHandle, checker: TsCheckerHandle) => ExtensionPayloadValue;
```

Then export a `defineCustomTypeWithTS<P>({ resolvePayload: (t: ts.Type, c: ts.TypeChecker) => P, ... })` helper from `@formspec/build` (which already depends on `typescript`) that narrows the handles at the boundary. The brand is invisible in autocomplete and enforces a deliberate cast at one place rather than sprinkled through every extension.

### 4. Contract is silent on load-bearing invariants
**Location:** TSDoc on `resolvePayload`.

For a callback consumed by a static-analysis pipeline, the TSDoc should state:
- **Purity** — must be a pure function of `(type, checker)`; no I/O, no shared state.
- **Multiplicity** — "may be invoked multiple times for the same type"; no memoization is provided. Today, duplicate field uses of the same custom type will re-invoke it.
- **JSON-serializability** — the return type nominally enforces `ExtensionPayloadValue`, but nothing stops an author from casting and returning a live `ts.Type` or `ts.Symbol`. If the IR is ever cached (incremental builds, language-server state, already implied by the `@formspec/ts-plugin` trajectory), that payload corrupts the cache. Document it, and consider a runtime JSON-round-trip assertion at `class-analyzer.ts:1710` behind a DEBUG flag.
- **Error semantics** — whatever you choose for #1, document it.
- **`undefined` vs `null`** — the return type is `ExtensionPayloadValue` (no `undefined`), but the call site does `?? null`, so bare `return;` is silently coerced to `null`, indistinguishable from "no payload." Either tighten the return type (`ExtensionPayloadValue | undefined` and drop the `??`) or document the coercion.
- An `@example` block with the canonical two-line cast idiom.

### 5. `@formspec/core` boundary erosion
**Location:** `packages/core/src/extensions/index.ts`.

`@formspec/core` is deliberately framework-neutral; it has no `typescript` dep. After this PR, its public contract includes a callback whose arguments are only meaningful if you hold `ts.Type` / `ts.TypeChecker`. The claim "framework-neutral" is now partly fiction: the types are erased but the contract is TS-compiler-coupled. Pick a side and document it in `ARCHITECTURE.md`:

- **(a)** Accept the coupling. State explicitly: "extension callbacks reference the host TypeScript compiler by contract; core does not import `typescript` so the parameter types are erased to `unknown`."
- **(b)** Move the analysis-time callback to `@formspec/build` via a separate registration overlay (e.g., `registerCustomTypeAnalyzer({ typeId, extractPayload })`). `@formspec/core` stays purely structural (only `toJsonSchema` + matching fields); `@formspec/build` owns TS-observing hooks. Worth weighing — (b) keeps core honest and makes the asymmetry explicit via package boundary rather than prose.

### 6. Undocumented invariant: `type`/`checker` come from the host program
**Location:** `class-analyzer.ts:1710` + TSDoc.

`resolvePayload` receives the real host `ts.TypeChecker`. But PR #297 and #294 are trending toward a parallel synthetic TS program. If a future refactor routes custom-type resolution through the synthetic pipeline, `resolvePayload` gets a different program's `ts.Type`. Property lookups by name still work, but symbol identity, declaration files, and `checker.getTypeOfSymbol` results silently diverge. Add a one-line invariant comment at the construction site, and state in the TSDoc: *"`type` and `checker` refer to the host program. Extensions may rely on host-program symbol identity."* Gives the next refactor a deliberate tripwire.

### 7. Release tier may be too aggressive
**Location:** `packages/core/api-report/core.public.api.md`.

The PR ships `resolvePayload` at `@public`, but the test uses `defineCustomType` imported from `@formspec/core/internals`, signaling the surrounding surface is still stabilizing. Consider marking just this member `@alpha` until at least one in-tree extension consumes it; core can stay `@public` otherwise. Cheaper to evolve the shape (especially if you take #3 and introduce branded handles).

## Test gaps

### 8. Assertions are tautological
**Location:** `packages/build/src/__tests__/resolve-payload.test.ts:89-93, 181-185`.

`toJsonSchema` is a `vi.fn()` the test authors to return exactly the shape being asserted. `expect(toHaveBeenCalledWith("customer", ...))` (line 86) is the only load-bearing assertion — the `toMatchObject({ "x-ref-target": "customer" })` check is just verifying the mock's return round-trips through the generator. Add **one** assertion that the payload lands on the canonical IR node directly (the existing `extension-runtime.integration.test.ts:97-103` pattern shows how). That proves the payload reached the IR, not just the mock.

### 9. Missing coverage
- **Throwing callback**: what's the contract, and does it crash or get attributed? Pin it.
- **Invocation multiplicity**: fixture with two fields using `TestRef<"customer">` + `toHaveBeenCalledTimes(N)` — documents the no-memoization contract.
- **`brand`-matched types**: current tests use `tsTypeNames` matching. The `brand` path is part of `CustomTypeRegistration`'s public contract; verify it also invokes `resolvePayload`, or file a follow-up.
- **Return `undefined` (vs `null`)**: pins the coercion in #4.
- **Non-serializable return (`new Date()`, `ts.Type`)**: document the failure mode, even if it's just "silently included until IR serialization fails."

## Minors / nits

- **Changeset scope**: `minor` across every package. Only `@formspec/core` has a new API surface; dependents re-export but add nothing. Larger version blast radius than behavior change. Confirm against repo policy.
- **Options-object future-proofing**: positional `(type, checker)` is fine today, but if a `sourceNode` / `diagnostics` context is plausible later, `(ctx: { type, checker })` is extensible non-breakingly. Worth considering now.
- **Symmetric generic** (speculative): `interface CustomTypeRegistration<P extends ExtensionPayloadValue = ExtensionPayloadValue>` with `resolvePayload?: (...) => P` and `toJsonSchema: (payload: P, ...) => ...` would give each registration a precise payload shape end-to-end. Catches mismatches at registration time. Default parameter preserves back-compat. Nice-to-have, not required.
- **Analysis vs. generation asymmetry is now a principle** worth stating once in `ARCHITECTURE.md`: analysis-time may observe the host program; generation-time is deterministic from the IR. Future PRs get something to point at.
- **Precedent pressure**: the next request will be analogous hooks for enums / discriminated unions / arrays. Don't pre-generalize, but raise the open question in ARCHITECTURE.md so the next author doesn't re-litigate it fresh.

## Strengths

- Single construction site in `resolveTypeNode` — verified by grep, no stale `payload: null` elsewhere.
- `ExtensionPayloadValue` as the return type is the right compile-time shape.
- `?? null` correctly preserves falsy-but-valid JSON values (`0`, `false`, `""`).
- Backwards-compatible additive change on an optional `readonly` field.
- API Extractor reports updated across all three release tiers.
- PR description is clear on the `unknown`-param rationale.

## Recommendation

**Comment / soft request-changes.** The strategic direction is correct and the gap is real. Before merge, I'd like to see: (1) error attribution or a documented fail-fast contract, (2) naming decision on `resolvePayload` vs `extractPayload`, (3) contract tightening in the TSDoc (purity, multiplicity, JSON-serializability, error semantics, `undefined` vs `null`), (4) one IR-level test assertion to escape tautology, (5) a test for the throw case. The opaque-handles refactor (#3) and the core/build boundary decision (#5) can land as a follow-up, but the outcome should be documented in ARCHITECTURE.md before a second analysis-time callback arrives.

PR #308, opened 2026-04-19
fix(build): prevent stack overflow on large external type arguments

brooks-stripe:
## Problem

`Ref<Stripe.Customer>` in a custom object field definition causes a stack overflow during schema generation. `Stripe.Customer` has 100+ deeply nested properties, and `extractReferenceTypeArguments` recursively calls `resolveTypeNode` on the type argument, which expands the full property tree and overflows the stack.

## Fix

When resolving type arguments in `extractReferenceTypeArguments`, if the argument is a named type declared in an external module (not the current file), emit it as an opaque reference node instead of recursively expanding it. Type arguments are only used for `$defs` naming and reference node `typeArguments` arrays — they don't contribute to the parent type's property schema. The parent type's properties (e.g., `id` and `type` on `Ref`) are resolved separately in the property loop, where `__` prefixed members like `__type?: T` are already skipped by `shouldEmitResolvedObjectProperty`.

This preserves the existing discriminator/prefix pipeline (`@discriminator :type T` + `apiNamePrefix`) that handles `Ref<OtherCustomObject>` correctly.

## Test plan

- [x] New: schema generation completes without overflow when a type argument is a large external type (100+ properties)
- [x] All 770 build tests pass
- [x] Manually tested: `pnpm image` in demo app produces `OBJECT_REFERENCE_TYPE` with `referencedType: "customer"` for `Ref<Customer>` using `@stripe/extensibility-api-objects`

### DIFF
--- a/.changeset/fix-large-type-argument-overflow.md
+++ b/.changeset/fix-large-type-argument-overflow.md
@@ -1,0 +1,8 @@
+---
+"@formspec/build": patch
+"@formspec/cli": patch
+"@formspec/eslint-plugin": patch
+"formspec": patch
+---
+
+Fix stack overflow when a generic type's type argument references a large external type (e.g., `Ref<Stripe.Customer>` where `Customer` has 100+ nested properties). Type arguments from external modules are now emitted as opaque references instead of being recursively expanded, since they are only used for `$defs` naming and don't contribute to the schema output.
--- a/packages/build/src/__tests__/large-type-argument.test.ts
+++ b/packages/build/src/__tests__/large-type-argument.test.ts
@@ -1,0 +1,93 @@
+/**
+ * Tests that type arguments referencing large external types don't overflow
+ * the stack during schema generation.
+ *
+ * Regression: Ref<Stripe.Customer> caused a stack overflow because
+ * extractReferenceTypeArguments recursed into Customer's 100+ deeply
+ * nested properties. The type argument is only used for naming and $defs
+ * references — it doesn't need full property expansion.
+ */
+
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
+import { generateSchemas, type GenerateSchemasOptions } from "../generators/class-schema.js";
+
+function generateSchemasOrThrow(options: Omit<GenerateSchemasOptions, "errorReporting">) {
+  return generateSchemas({
+    ...options,
+    errorReporting: "throw",
+  });
+}
+
+let tmpDir: string;
+
+beforeAll(() => {
+  tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "formspec-large-type-arg-"));
+
+  // A large external type with many properties (simulates Stripe.Customer)
+  fs.writeFileSync(
+    path.join(tmpDir, "large-type.ts"),
+    [
+      "export interface LargeType {",
+      "  readonly object: 'large_thing';",
+      ...Array.from({ length: 100 }, (_, i) => `  prop${i}: string;`),
+      "}",
+    ].join("\n")
+  );
+
+  // A generic wrapper type that references the large type as a type argument
+  fs.writeFileSync(
+    path.join(tmpDir, "wrapper.ts"),
+    [
+      'import type { LargeType } from "./large-type.js";',
+      "",
+      "type Wrapper<T extends { readonly object: string }> = {",
+      "  id: string;",
+      '  type: T["object"];',
+      "};",
+      "",
+      "export interface Config {",
+      "  ref: Wrapper<LargeType>;",
+      "}",
+    ].join("\n")
+  );
+
+  fs.writeFileSync(
+    path.join(tmpDir, "tsconfig.json"),
+    JSON.stringify(
+      {
+        compilerOptions: {
+          target: "ES2022",
+          module: "NodeNext",
+          moduleResolution: "nodenext",
+          strict: true,
+          skipLibCheck: true,
+        },
+      },
+      null,
+      2
+    )
+  );
+});
+
+afterAll(() => {
+  if (tmpDir) {
+    fs.rmSync(tmpDir, { recursive: true, force: true });
+  }
+});
+
+describe("large external type argument", () => {
+  it("does not overflow the stack when a type argument is a large external type", () => {
+    const result = generateSchemasOrThrow({
+      filePath: path.join(tmpDir, "wrapper.ts"),
+      typeName: "Config",
+    });
+
+    // The build should complete without stack overflow. The wrapper's own
+    // properties (id, type) should appear in the schema output.
+    expect(result.jsonSchema).toBeDefined();
+    expect(result.jsonSchema.properties).toBeDefined();
+  });
+});
--- a/packages/build/src/__tests__/large-type-argument.test.ts
+++ b/packages/build/src/__tests__/large-type-argument.test.ts
@@ -2,14 +2,15 @@
  * Tests for generic wrapper types (like Ref<T>) with type arguments of
  * varying sizes and origins.
  *
+ * Related: packages/build/src/__tests__/discriminator.test.ts:704 covers
+ * a similar large-carrier scenario via @discriminator.
+ *
  * Covers:
- *   1. Small locally-defined type argument — resolves correctly with
- *      the wrapper's own properties preserved.
- *   2. Deeply nested external type argument — doesn't overflow the stack
- *      and doesn't leak into $defs.
- *      Regression: Ref<Stripe.Customer> caused a stack overflow because
- *      extractReferenceTypeArguments recursed into Customer's deeply
- *      nested property tree.
+ *   1. External deeply-nested type argument — doesn't overflow the stack
+ *      and doesn't leak into $defs (regression for Ref<Stripe.Customer>).
+ *   2. External small type argument — same opaque-reference behavior.
+ *   3. Same-file type argument — falls through to normal recursive resolution.
+ *   4. Inline anonymous type argument — falls through to normal resolution.
  */
 
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
@@ -40,27 +41,18 @@ beforeAll(() => {
     ].join("\n")
   );
 
-  // Deeply nested type chain (simulates Stripe.Customer's recursive depth).
-  // Each level references the next, creating a chain that would reliably
-  // overflow the stack if recursively expanded.
-  const nestingDepth = 30;
-  const nestedTypeLines: string[] = [
-    "export interface Level0 {",
-    "  readonly object: 'deep_thing';",
-    "  name: string;",
-    "  nested: Level1;",
-    "}",
-  ];
-  for (let i = 1; i < nestingDepth; i++) {
-    const next = i < nestingDepth - 1 ? `Level${String(i + 1)}` : "string";
-    nestedTypeLines.push(
-      `export interface Level${String(i)} {`,
-      `  value${String(i)}: string;`,
-      `  nested: ${next};`,
-      "}",
-    );
+  // Deeply nested type chain (simulates the deeply-recursive shape that caused
+  // the original Ref<Stripe.Customer> overflow). Each type references the next,
+  // so resolving one requires recursing through all ~300 declarations.
+  const depth = 300;
+  const lines: string[] = [];
+  for (let d = 0; d < depth; d++) {
+    lines.push(`export interface Type${String(d)} {`);
+    lines.push(`  readonly object: 'type_${String(d)}';`);
+    if (d + 1 < depth) lines.push(`  nested: Type${String(d + 1)};`);
+    lines.push("}");
   }
-  fs.writeFileSync(path.join(tmpDir, "deep-type.ts"), nestedTypeLines.join("\n"));
+  fs.writeFileSync(path.join(tmpDir, "deep-type.ts"), lines.join("\n"));
 
   // Generic wrapper (simulates Ref<T>)
   fs.writeFileSync(
@@ -75,7 +67,7 @@ beforeAll(() => {
     ].join("\n")
   );
 
-  // Fixture using small local type argument
+  // Fixture using small external type argument
   fs.writeFileSync(
     path.join(tmpDir, "small-ref.ts"),
     [
@@ -92,11 +84,46 @@ beforeAll(() => {
   fs.writeFileSync(
     path.join(tmpDir, "deep-ref.ts"),
     [
-      'import type { Level0 } from "./deep-type.js";',
+      'import type { Type0 } from "./deep-type.js";',
       'import type { Wrapper } from "./wrapper.js";',
       "",
       "export interface DeepRefConfig {",
-      "  ref: Wrapper<Level0>;",
+      "  ref: Wrapper<Type0>;",
+      "}",
+    ].join("\n")
+  );
+
+  // Fixture using same-file type argument (should resolve normally)
+  fs.writeFileSync(
+    path.join(tmpDir, "same-file-ref.ts"),
+    [
+      "type Wrapper<T extends { readonly object: string }> = {",
+      "  id: string;",
+      '  type: T["object"];',
+      "};",
+      "",
+      "interface LocalType {",
+      "  readonly object: 'local_thing';",
+      "  extra: number;",
+      "}",
+      "",
+      "export interface SameFileConfig {",
+      "  ref: Wrapper<LocalType>;",
+      "}",
+    ].join("\n")
+  );
+
+  // Fixture using inline anonymous type argument (should resolve normally)
+  fs.writeFileSync(
+    path.join(tmpDir, "inline-ref.ts"),
+    [
+      "type Wrapper<T extends { readonly object: string }> = {",
+      "  id: string;",
+      '  type: T["object"];',
+      "};",
+      "",
+      "export interface InlineConfig {",
+      "  ref: Wrapper<{ readonly object: 'inline_thing' }>;",
       "}",
     ].join("\n")
   );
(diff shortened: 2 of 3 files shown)


mike-north:
@codex review this

.changeset/fix-large-type-argument-overflow.md
```diff
@@ -0,0 +1,8 @@
+---
+"@formspec/build": patch
+"@formspec/cli": patch
+"@formspec/eslint-plugin": patch
+"formspec": patch
```
mike-north, line 5: Process note (no suggestion): this branch is one commit behind `origin/main`. PR #305 (`feat(analysis): structured constraint-validator logging`) was merged after this PR opened and modified the same file (`packages/build/src/analyzer/class-analyzer.ts`) — adding `logExtractPayload` calls in `resolveTypeNode`, just below `extractReferenceTypeArguments`. A `git merge-tree` check shows no textual conflict, but please rebase onto current `main` and re-push before merging so reviewers can evaluate this change against the latest state of the file.

packages/build/src/__tests__/large-type-argument.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    const refSchema = properties["ref"] as Record<string, unknown>;
+    expect(refSchema).toBeDefined();
+  });
+
+  it("resolves Wrapper<Level0> without stack overflow and without leaking into $defs", () => {
+    const result = generateSchemasOrThrow({
+      filePath: path.join(tmpDir, "deep-ref.ts"),
+      typeName: "DeepRefConfig",
+    });
+
```
mike-north, line 192: **Both tests are tautological** — I verified by reverting `packages/build/src/analyzer/class-analyzer.ts` to its pre-fix state: 2/2 of these tests still pass. `expect(refSchema).toBeDefined()` is satisfied by any non-undefined value, including empty objects. Per the repo's bug-fix regression rule, the test must *fail* against the pre-fix code to guard the behavior this PR introduces.

The observable difference the fix creates (for `Ref<LargeType>` with the current fixture):

- Pre-fix: `$defs` contains BOTH `Wrapper__LargeType` AND `LargeType` (with all 100 `prop*` fields expanded).
- Post-fix: `$defs` contains ONLY `Wrapper__LargeType`; the external type is opaque.

Replace the whole `describe` block with structural assertions that would fail on pre-fix and also exercise the `@discriminator :type T` + `T["object"]` pipeline that the PR body claims is preserved:

```suggestion
  it("regression: Wrapper<SmallType> keeps wrapper's own properties and does NOT inline the external type argument's properties", () => {
    const root = result.jsonSchema as Record<string, unknown>;
    const defs = (root["$defs"] ?? {}) as Record<string, unknown>;
    const props = root["properties"] as Record<string, unknown>;
    const ref = props["ref"] as Record<string, unknown>;
    expect(ref["$ref"]).toBe("#/$defs/Wrapper__SmallType");

    const wrapperDef = defs["Wrapper__SmallType"] as Record<string, unknown>;
    const wrapperProps = wrapperDef["properties"] as Record<string, unknown>;
    expect(Object.keys(wrapperProps).sort()).toEqual(["id", "type"]);
    expect((wrapperProps["id"] as Record<string, unknown>)["type"]).toBe("string");
    // Discriminator pipeline still resolves T["object"] for external T.
    expect((wrapperProps["type"] as Record<string, unknown>)["enum"]).toEqual(["small_thing"]);

    // Pre-fix, $defs also contained "SmallType" with its properties inlined.
    expect(defs).not.toHaveProperty("SmallType");
  it("regression: Wrapper<Type0> emits deeply-nested external type as opaque reference without recursing through the chain", () => {
    const root = result.jsonSchema as Record<string, unknown>;
    const defs = (root["$defs"] ?? {}) as Record<string, unknown>;
    const props = root["properties"] as Record<string, unknown>;

    expect((props["ref"] as Record<string, unknown>)["$ref"]).toBe("#/$defs/Wrapper__Type0");

    const wrapperDef = defs["Wrapper__Type0"] as Record<string, unknown>;
    const wrapperProps = wrapperDef["properties"] as Record<string, unknown>;
    expect(Object.keys(wrapperProps).sort()).toEqual(["id", "type"]);
    expect((wrapperProps["type"] as Record<string, unknown>)["enum"]).toEqual(["type_0"]);
    // Pre-fix, $defs.Type0 (and all nested Type1..Type299) would have been
    // inlined with their properties — and the recursion caused a stack overflow.
    expect(defs).not.toHaveProperty("Type0");
    expect(wrapperProps).not.toHaveProperty("nested");
```

mike-north, line 192: The fix's new block has two untested branches:

1. **Same-file type argument** (the `else` that falls through to the original recursive `resolveTypeNode` call). Both current fixtures import `T` from a separate file. A future refactor that inverts the condition or drops the else would silently break same-file `Wrapper<LocalType>` expansion with no test failure.
2. **Anonymous/inline type argument** (`argumentName === undefined` fallback). `Wrapper<{ object: 'inline' }>` takes the original recursive path because the inline object has no symbol name. This is currently uncovered.

Add both cases. Illustrative suggestion (insert before the closing `});`):

```suggestion
  it("regression: Wrapper<SmallType> keeps wrapper's own properties and does NOT inline the external type argument's properties", () => {
    const root = result.jsonSchema as Record<string, unknown>;
    const defs = (root["$defs"] ?? {}) as Record<string, unknown>;
    const props = root["properties"] as Record<string, unknown>;
    const ref = props["ref"] as Record<string, unknown>;
    expect(ref["$ref"]).toBe("#/$defs/Wrapper__SmallType");

    const wrapperDef = defs["Wrapper__SmallType"] as Record<string, unknown>;
    const wrapperProps = wrapperDef["properties"] as Record<string, unknown>;
    expect(Object.keys(wrapperProps).sort()).toEqual(["id", "type"]);
    expect((wrapperProps["id"] as Record<string, unknown>)["type"]).toBe("string");
    // Discriminator pipeline still resolves T["object"] for external T.
    expect((wrapperProps["type"] as Record<string, unknown>)["enum"]).toEqual(["small_thing"]);

    // Pre-fix, $defs also contained "SmallType" with its properties inlined.
    expect(defs).not.toHaveProperty("SmallType");
  it("regression: Wrapper<Type0> emits deeply-nested external type as opaque reference without recursing through the chain", () => {
    const root = result.jsonSchema as Record<string, unknown>;
    const defs = (root["$defs"] ?? {}) as Record<string, unknown>;
    const props = root["properties"] as Record<string, unknown>;

    expect((props["ref"] as Record<string, unknown>)["$ref"]).toBe("#/$defs/Wrapper__Type0");

    const wrapperDef = defs["Wrapper__Type0"] as Record<string, unknown>;
    const wrapperProps = wrapperDef["properties"] as Record<string, unknown>;
    expect(Object.keys(wrapperProps).sort()).toEqual(["id", "type"]);
    expect((wrapperProps["type"] as Record<string, unknown>)["enum"]).toEqual(["type_0"]);
    // Pre-fix, $defs.Type0 (and all nested Type1..Type299) would have been
    // inlined with their properties — and the recursion caused a stack overflow.
    expect(defs).not.toHaveProperty("Type0");
    expect(wrapperProps).not.toHaveProperty("nested");
```

packages/build/src/__tests__/large-type-argument.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  );
+
+  // Deeply nested type chain (simulates Stripe.Customer's recursive depth).
+  // Each level references the next, creating a chain that would reliably
+  // overflow the stack if recursively expanded.
+  const nestingDepth = 30;
+  const nestedTypeLines: string[] = [
+    "export interface Level0 {",
+    "  readonly object: 'deep_thing';",
+    "  name: string;",
```
mike-north: The 100-flat-property fixture does not actually reproduce the original stack overflow — I verified that reverting the analyzer fix and running this test still passes. The real overflow requires a *deeply nested* type chain (e.g., `Type0 { nested: Type1 }; Type1 { nested: Type2 }; …`), which is what `Stripe.Customer` looks like under the hood. Without a deeper fixture this test is a "completion without throw" smoke test, not a regression guard for the advertised scenario.

Then update the consumer fixture (`large-ref.ts`) to reference `Type0` (and rename `LargeType` → `Type0` throughout). The test becomes a genuine regression check: it throws on pre-fix code and passes on post-fix code.

```suggestion
  // Large nested type chain (simulates the deeply-recursive shape that caused
  // the original Ref<Stripe.Customer> overflow). Each type references the next,
  // so resolving one requires recursing through all ~300 declarations.
  const depth = 300;
  const lines: string[] = [];
  for (let d = 0; d < depth; d++) {
    lines.push(`export interface Type${String(d)} {`);
    lines.push(`  readonly object: 'type_${String(d)}';`);
    if (d + 1 < depth) lines.push(`  nested: Type${String(d + 1)};`);
    lines.push("}");
  }
  fs.writeFileSync(path.join(tmpDir, "large-type.ts"), lines.join("\n"));
```

packages/build/src/__tests__/large-type-argument.test.ts
```diff
@@ -0,0 +1,159 @@
...
+beforeAll(() => {
+  tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "formspec-type-arg-"));
+
+  // Small type defined in a separate file
+  fs.writeFileSync(
+    path.join(tmpDir, "small-type.ts"),
+    [
+      "export interface SmallType {",
+      "  readonly object: 'small_thing';",
+      "}",
```
mike-north, line 40: `packages/build/src/__tests__/discriminator.test.ts:704` already has a regression test for a closely related code path ("specializes Ref<T> for large object carriers without expanding the target object") using a 101-field carrier. The new file is a legitimately distinct case (phantom `T["object"]` index access instead of `@discriminator`), but the overlap is invisible in a standalone file. Either:

- Add a cross-reference to the file-level comment pointing to `discriminator.test.ts:704`, OR
- Fold the new case into `discriminator.test.ts` as a second `describe` block so the tmpdir/tsconfig/carrier boilerplate is shared.

Either option makes it easier for future maintainers to find all the large-type regression coverage in one place. Additionally: `generateSchemasOrThrow` is now copy-pasted verbatim across 10 test files in this package — a shared `__tests__/helpers/generate-schemas.ts` is good follow-up work (not a blocker for this PR).

packages/build/src/analyzer/class-analyzer.ts
```diff
@@ -1292,6 +1292,38 @@ function extractReferenceTypeArguments(
...
+    // Resolve through import aliases to the canonical declaration so that
+    // the source-file comparison checks the external module, not the local
+    // import specifier.
+    const argumentSymbol =
+      rawSymbol !== undefined && (rawSymbol.flags & ts.SymbolFlags.Alias) !== 0
+        ? checker.getAliasedSymbol(rawSymbol)
+        : rawSymbol;
+    const argumentDecl = argumentSymbol?.declarations?.[0];
+    if (
+      argumentDecl !== undefined &&
```
mike-north, line 1312: The heuristic `argumentDecl.getSourceFile().fileName !== sourceNode?.getSourceFile().fileName` compares against the *immediate call-site's* source file, not the canonical analysis root. Two concrete problems:

1. **Contradicts the comment.** The comment (lines 1296-1300) says "named type from an external module," but this check fires for any sibling `.ts` file in the same package. I verified empirically: for `Ref<SmallType>` where `SmallType` lives in `./small-type.ts`, pre-fix emits `$defs.SmallType`; post-fix does not. This is a real behavior change beyond npm-packaged types.
2. **Dead-defensive optional chain.** `sourceNode?.getSourceFile().fileName` can theoretically produce `string | undefined`, but the earlier guard at lines 1275-1277 already returns `[]` when `sourceTypeNode` is undefined (which in turn is defined iff `sourceNode` is). If that earlier guard is ever relaxed, `"x.ts" !== undefined` silently forces *every* type argument through the opaque path.

The function already receives `file: string` — the canonical analysis root. Line 2428 in this same file uses that parameter for the same kind of check (`namedDecl?.getSourceFile().fileName !== file`). Align with that convention:

```suggestion
      argumentDecl.getSourceFile().fileName !== file
```

packages/build/src/analyzer/class-analyzer.ts
```diff
@@ -1292,6 +1292,38 @@ function extractReferenceTypeArguments(
...
+    // overflows on deeply nested external types (e.g. Stripe.Customer)
+    // that are only used as phantom type parameters (e.g. Ref<T>).
+    const rawSymbol = argumentType.aliasSymbol ?? argumentType.getSymbol();
+    // Resolve through import aliases to the canonical declaration so that
+    // the source-file comparison checks the external module, not the local
+    // import specifier.
+    const argumentSymbol =
+      rawSymbol !== undefined && (rawSymbol.flags & ts.SymbolFlags.Alias) !== 0
+        ? checker.getAliasedSymbol(rawSymbol)
+        : rawSymbol;
```
mike-north: `as const` on `kind` and `as TypeNode[]` on `typeArguments` are both unnecessary here and inconsistent with the 7 other `{ kind: "reference", ..., typeArguments: [] }` literals in this file (lines 1989, 2084, 2143, 2441, 2474, 2531, 2669). The `.map()` callback is contextually typed to `{ readonly tsType: ts.Type; readonly typeNode: TypeNode }` via the function's declared return type, so the literal narrows to `ReferenceTypeNode` without either cast.

Also: `[] as TypeNode[]` widens to a *mutable* array, while `ReferenceTypeNode.typeArguments` is declared `readonly TypeNode[]` in `packages/core/src/types/ir.ts:257`. The assignment happens to compile (mutable→readonly is assignable), but it diverges from the six sibling sites that use the bare literal and preserve the readonly alignment.

```suggestion
          typeNode: {
            kind: "reference",
            name: argumentName,
            typeArguments: [],
          },
```

packages/build/src/analyzer/class-analyzer.ts
```diff
@@ -1292,6 +1292,38 @@ function extractReferenceTypeArguments(
...
     const argumentType = checker.getTypeFromTypeNode(argumentNode);
+
+    // If the type argument is a named type from an external module (not
+    // declared in the current file), register it as an opaque reference
+    // rather than resolving its full property tree. This prevents stack
+    // overflows on deeply nested external types (e.g. Stripe.Customer)
+    // that are only used as phantom type parameters (e.g. Ref<T>).
+    const rawSymbol = argumentType.aliasSymbol ?? argumentType.getSymbol();
+    // Resolve through import aliases to the canonical declaration so that
+    // the source-file comparison checks the external module, not the local
```
mike-north: Two inaccuracies here that will mislead future maintainers:

1. **"register it as an opaque reference"** — no registration occurs. The code returns a synthetic `TypeNode` without touching `typeRegistry`. Downstream `$refs` rely on the parent type's property walk (or `buildInstantiatedReferenceName`), not on this branch.
2. **"from an external module"** — the check is file-level, not module-level. Any sibling file in the same package hits this path.

```suggestion
    // If the type argument is a named type declared outside the current
    // analysis root file, emit a synthetic opaque reference `TypeNode`
    // (`{ kind: "reference", name, typeArguments: [] }`) instead of recursing
    // into its declaration. This preserves the name for `buildInstantiatedReferenceName`
    // while avoiding stack overflows on deeply-nested types (e.g. Stripe.Customer)
    // used as phantom parameters (e.g. Ref<T>). Note: this path does NOT populate
    // typeRegistry — callers that need a resolved $defs entry must get it via the
    // parent type's property walk.
```

packages/build/src/analyzer/class-analyzer.ts
```diff
@@ -1292,6 +1292,38 @@ function extractReferenceTypeArguments(
...
+    // import specifier.
+    const argumentSymbol =
+      rawSymbol !== undefined && (rawSymbol.flags & ts.SymbolFlags.Alias) !== 0
+        ? checker.getAliasedSymbol(rawSymbol)
```
mike-north: For robustness on the advertised `Stripe.Customer` scenario: this file already has an established pattern for chasing symbol aliases at line 1064 —

```typescript
const symbol = type.aliasSymbol ?? type.getSymbol();
const aliased =
  symbol !== undefined && (symbol.flags & ts.SymbolFlags.Alias)
    ? checker.getAliasedSymbol(symbol)
    : undefined;
const targetSymbol = aliased ?? symbol;
```

The new block stops at `argumentSymbol?.declarations?.[0]` without chasing import aliases. For ordinary interface imports this happens to work (because `getTypeFromTypeNode` already resolves aliases into the declaring symbol), but for namespace-qualified types like `Stripe.Customer` — the exact scenario this PR claims to fix — the symbol chain can differ. Mirroring the line-1064 pattern costs ~4 lines and matches existing internal convention.

```suggestion
    const baseSymbol = argumentType.aliasSymbol ?? argumentType.getSymbol();
    const argumentSymbol =
      baseSymbol !== undefined && baseSymbol.flags & ts.SymbolFlags.Alias
        ? checker.getAliasedSymbol(baseSymbol)
        : baseSymbol;
```

packages/build/src/analyzer/class-analyzer.ts
mike-north: Minor design suggestion (no strict mechanical replacement): consider extracting the opaque-reference construction into a small helper so the intent is documented at its declaration and future IR shape changes have a single touchpoint:

```typescript
function makeOpaqueReferenceTypeNode(name: string): TypeNode {
  // typeArguments is intentionally empty: this type is used only for $defs
  // naming, not for schema output. Callers must not rely on the arg list.
  return { kind: "reference", name, typeArguments: [] };
}
```

Then the branch inside `.map()` becomes:

```typescript
return { tsType: argumentType, typeNode: makeOpaqueReferenceTypeNode(argumentName) };
```

Debatable for a single call-site — if you prefer the inline form, softening the comment above to explain *why* `typeArguments: []` is safe accomplishes the same goal.

mike-north (approved):

PR #313, opened 2026-04-19
chore(core): remove extractPayload from CustomTypeRegistration

brooks-stripe:
## Why?

`extractPayload` was added in #300 to support `Ref<T>` by extracting the referenced type tag from the generic argument at analysis time. This approach was superseded by #308, which fixes the underlying stack overflow in `extractReferenceTypeArguments` — allowing formspec's existing object resolution and discriminator pipeline to handle `Ref<T>` correctly without custom type registration.

## What?

- Remove `extractPayload` field from `CustomTypeRegistration` in `@formspec/core`
- Remove the `extractPayload` call site in `class-analyzer.ts`
- Remove `resolve-payload.test.ts`

## Test plan

- [x] All 764 build tests pass
- [x] No in-tree consumers of `extractPayload`


mike-north:
@codex review this

packages/build/src/analyzer/class-analyzer.ts
```diff
@@ -1707,32 +1706,10 @@ export function resolveTypeNode(
...
-      customTypeName: customTypeLookup.registration.typeName,
-      tsApisTouched,
-    });
     return {
       kind: "custom",
-      typeId,
-      payload,
+      typeId: customTypeIdFromLookup(customTypeLookup),
+      payload: null,
     };
```
mike-north, line 1713: After this PR, `logExtractPayload` and `ExtractPayloadLogEntry` in `@formspec/analysis` are dead code — their only call site was the one removed from this file. They are still exported from the internal surface:

- `packages/analysis/src/constraint-validator-logger.ts:120-145` — `ExtractPayloadLogEntry` interface
- `packages/analysis/src/constraint-validator-logger.ts:395-406` — `logExtractPayload` function
- `packages/analysis/src/internal.ts:187` — `export { logExtractPayload }`
- `packages/analysis/src/internal.ts:192` — `export type { ExtractPayloadLogEntry }`

Four reviewers independently flagged this. Retiring them atomically in this PR keeps the revert clean and avoids leaving an `@formspec/analysis/internal` surface whose only purpose was the feature this PR removes. If you'd rather keep this PR narrow, please file a follow-up issue before merging so the cleanup isn't lost.

There are also four stale references to `extractPayload` in `docs/refactors/synthetic-checker-retirement.md` (lines 180, 216, 256, 268) that would no longer accurately describe the codebase after this PR — worth updating in the same sweep.

packages/build/src/analyzer/class-analyzer.ts
```diff
@@ -1707,32 +1706,10 @@ export function resolveTypeNode(
...
-      extensionId: customTypeLookup.extensionId,
-      customTypeName: customTypeLookup.registration.typeName,
-      tsApisTouched,
-    });
     return {
       kind: "custom",
-      typeId,
-      payload,
+      typeId: customTypeIdFromLookup(customTypeLookup),
+      payload: null,
```
mike-north, line 1712: Heads-up on a latent conflict: PR #319 (`fix/309-enum-defs-deduplication` by @mike-north, currently open) also modifies `packages/build/src/analyzer/class-analyzer.ts`. Whichever lands second will need to rebase, and given how small this hunk is the merge should be trivial — but worth knowing before you queue merges.

packages/build/src/analyzer/class-analyzer.ts
mike-north, line 1713: With `extractPayload` gone, `payload` on this node is produced as `null` at every call site, and `CustomTypeRegistration.toJsonSchema`'s first parameter is now always `null`. The `payload` field on `CustomTypeNode` (in `packages/core/src/types/ir.ts:284-298`) and the first parameter of `toJsonSchema` are therefore vestigial — the type system still allows non-null values that can never occur in practice.

Out of scope for this PR (the type is `@beta` and the simplification is another breaking change), but worth capturing as a follow-on so the IR contract reflects reality: either drop `payload` from `CustomTypeNode` and from `toJsonSchema`'s signature, or leave a comment noting the intentional extension point if the plan is to restore it later.

brooks-stripe: Resolved by 2df6c136e7f5f65ac19b9c132780ce4ab403863b.

PR #317, opened 2026-04-19
test: pin three known build/snapshot divergences (0.5f)

mike-north:
## Summary

Implements Phase 0.5f of the synthetic-checker retirement plan (§9.3 #16).

Adds `packages/analysis/src/__tests__/parity-divergences.test.ts` with **six tests** (two per divergent input) that pin the exact diagnostic output each consumer produces today.

### 1. `@const not-json`

| Consumer | Argument rendered as | Diagnostic |
|---|---|---|
| **Build** | `"not-json"` (JSON.stringify of raw text) | **none** — string satisfies `JsonValue = unknown` |
| **Snapshot** | *(omitted — JSON.parse throws → null)* | **`INVALID_TAG_ARGUMENT`**: `"Expected 2-3 arguments, but got 1."` |

Root cause: `renderSyntheticArgumentExpression` (build) falls back to `JSON.stringify(text)` for invalid JSON; `getArgumentExpression` (snapshot) returns `null`, omitting the argument from the synthetic call entirely.

### 2. `@minimum Infinity`

| Consumer | Argument rendered as | Diagnostic |
|---|---|---|
| **Build** | `"Infinity"` (JSON.stringify, since `!Number.isFinite(Infinity)`) | **`TYPE_MISMATCH`**: string not assignable to number |
| **Snapshot** | `Infinity` (identifier, passed through unchanged) | **none** — `Infinity` has type `number` in TypeScript |

### 3. `@minimum NaN`

| Consumer | Argument rendered as | Diagnostic |
|---|---|---|
| **Build** | `"NaN"` (JSON.stringify, since `!Number.isFinite(NaN)`) | **`TYPE_MISMATCH`**: string not assignable to number |
| **Snapshot** | `NaN` (identifier, passed through unchanged) | **none** — `NaN` has type `number` in TypeScript |

## Design

- **Build consumer tests** call `checkSyntheticTagApplication` directly with argument text rendered using the same logic as `renderSyntheticArgumentExpression` in `tsdoc-parser.ts` — no cross-package imports required.
- **Snapshot consumer tests** use `buildFormSpecAnalysisFileSnapshot` with inline TypeScript source strings.
- Each describe block carries a `// KNOWN DIVERGENCE (refactor plan §3)` comment documenting the divergence and its role as an anchor for the eventual normalization PR.
- Includes a `@formspec/analysis` patch changeset.


mike-north:
@codex please review this

mike-north:
@codex review this

mike-north:
Addressed in commit cd8861f9. Here's what was done and what was discovered:

**Approach taken:** The test file was moved from `packages/analysis` to `packages/build` so the build-side probe could call `generateSchemas()` (the real build entry point) rather than a local reimplementation of `renderSyntheticArgumentExpression`. `@formspec/analysis` has no dependency on `@formspec/build`, so the test had to move where both paths are accessible.

`runBuildConsumer` now calls:
```
generateSchemas({ filePath, typeName: "TestForm", errorReporting: "diagnostics" })
```
…which routes through `analyzeInterfaceToIR → parseTSDocTags → buildCompilerBackedConstraintDiagnostics → renderSyntheticArgumentExpression → checkSyntheticTagApplication → validateIR`. If `renderSyntheticArgumentExpression` changes, these tests will catch it.

**Behavior changes discovered by the real pipeline:**

1. **`@const not-json` BUILD case** — the original test assumed zero diagnostics. The real build pipeline _does_ emit `TYPE_MISMATCH`, but from the IR validator rather than the synthetic type checker: `"Field "value": @const value type "string" is incompatible with field type "number"`. The synthetic checker silently passes (`"not-json"` as string satisfies `JsonValue = unknown`), but the IR validator catches it because a string literal is not a valid `number` constant. The divergence with the snapshot path is preserved — it still emits `INVALID_TAG_ARGUMENT` at argument-count check time.

2. **`@minimum Infinity` / `@minimum NaN`** — both still produce `TYPE_MISMATCH`, but the message is now the stable FormSpec-level message (`'Tag "@minimum" received an invalid argument for number.'`) rather than the raw TypeScript `'not assignable'` wording that the old test pinned.

All 775 build tests and 315 analysis tests pass. The old file in `packages/analysis` is replaced with a `describe.skip` redirect explaining the move.

PR #320, opened 2026-04-19
test(analysis): setup-diagnostic emission-count stability (0.5h)

mike-north:
## Summary

Pins the current setup-diagnostic emission-count model before Phase 4 of the synthetic-checker retirement plan relocates setup diagnostics into an extension-registry construction pass (§9.3 #19).

**Observed emission model: per-batch / per-call**

- `UNSUPPORTED_CUSTOM_TYPE_OVERRIDE` and `SYNTHETIC_SETUP_FAILURE` are both thrown during synthetic-prelude construction (`collectExtensionCustomTypeNames`), before the LRU cache key is computed
- Because the throw exits `runBatchSyntheticCheck` before reaching the cache, **the LRU cache is bypassed entirely for setup failures**
- Each `buildFormSpecAnalysisFileSnapshot` call with a failing extension config independently re-triggers the throw and emits a fresh diagnostic
- N snapshot refreshes → N diagnostics, regardless of whether the extension config object is the same reference or freshly constructed

**What is pinned**

| Test | Assertion |
|------|-----------|
| Single-call baseline | 1 `UNSUPPORTED_CUSTOM_TYPE_OVERRIDE` per call |
| Three repeated refreshes | 3 total (one per call, cache does not deduplicate setup failures) |
| Fresh extension config object | Same per-call cost (+1 delta), confirming no cross-object deduplication |
| `SYNTHETIC_SETUP_FAILURE` variant | Same per-batch model for the second setup-error kind |

**Phase 4 risk addressed**

If Phase 4 moves setup-diagnostic emission into `createExtensionRegistry`, the per-call count may change (e.g., 0 if deduplication happens at construction; or >1 if the diagnostic is re-emitted per-field). Any such change will cause these tests to fail and require explicit sign-off — the tests carry inline comments explaining the expected failure direction.

## Test plan

- [x] `pnpm run build` passes
- [x] `pnpm --filter @formspec/analysis run test` passes — 319 tests (4 new)
- [x] `pnpm run lint` passes
- [x] All 4 new tests pass against current code; each asserts the per-batch model documented in §9.3 #19

### DIFF
--- a/.changeset/phase-0.5h-emission-count.md
+++ b/.changeset/phase-0.5h-emission-count.md
@@ -1,0 +1,11 @@
+---
+"@formspec/analysis": patch
+"@formspec/build": patch
+"@formspec/cli": patch
+"@formspec/eslint-plugin": patch
+"@formspec/language-server": patch
+"@formspec/ts-plugin": patch
+"formspec": patch
+---
+
+Pin setup-diagnostic emission-count stability (Phase 0.5h, §9.3 #19).
--- a/packages/analysis/src/__tests__/setup-diagnostic-emission-count.test.ts
+++ b/packages/analysis/src/__tests__/setup-diagnostic-emission-count.test.ts
@@ -24,9 +24,11 @@
  *
  * If Phase 4 deduplicates by moving validation into a registry object:
  *   - The per-call count might drop to 0 (if validation only runs at registry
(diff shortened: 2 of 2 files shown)


mike-north:
@codex review this

PR #351, opened 2026-04-20
fix(eslint-plugin): recognize extension-registered annotation tags (#350)

### LINKED ISSUE #350: ESLint tag-recognition rules should check extension annotations, not just constraints and metadata slots
### Summary

The `tag-recognition/no-unknown-tags` and `tag-recognition/tsdoc-comment-syntax` ESLint rules don't recognize tags registered as `annotations` in the extension registry. They only check `constraintTags` and `metadataSlots`.

### Context

The Stripe SDK defines `@primaryField` and `@secondaryField` as custom annotation tags for custom objects. These should be registered via `defineAnnotation()` in the extension definition. However, the ESLint rules flag them as unknown:

```
error  The TSDoc tag "@primaryField" is not defined in this configuration  formspec/tag-recognition/tsdoc-comment-syntax
```

### Root cause

In `tsdoc-comment-syntax.ts`, `readExtensionTagNames()` (lines 38-45) only iterates:
- `extension.constraintTags`
- `extension.metadataSlots`

It does not iterate `extension.annotations`.

Similarly, `no-unknown-tags` only checks `getTagDefinition()` from `@formspec/analysis` which only knows built-in tags — it doesn't check the extension registry at all.

### Fix

1. `readExtensionTagNames()` should also iterate `extension.annotations` (mapping `annotationName` to a tag name)
2. `no-unknown-tags` should also check extension-defined tags via `context.settings` (same mechanism `tsdoc-comment-syntax` uses)

### Related

Reported by shivn in #proj-custom-objects-app-orchestration

mike-north:
## Summary

- Fixes #350. `tag-recognition/no-unknown-tags` and `tag-recognition/tsdoc-comment-syntax` previously ignored tags registered as `annotations` on an extension (only `constraintTags` and `metadataSlots` were considered). As a result, tags declared via `defineAnnotation()` (e.g. `@primaryField`, `@secondaryField`) were reported as unknown.
- Extracts a shared `readExtensionTagNames()` helper in `packages/eslint-plugin/src/utils/tag-metadata.ts` that flattens `constraintTags + metadataSlots + annotations` from the settings-bound extension registry. `no-unknown-tags` now consults it via `context.settings`, matching how `tsdoc-comment-syntax` already does.
- Hardens the helper against malformed settings: non-object extension entries are skipped, and `tagName` / `annotationName` are string-guarded before use. Annotation names registered with a leading `@` are normalized.

Closes #350.

## Test plan

- [x] Failing regression tests added first (demonstrated the bug pre-fix), now passing:
  - Annotation-only declaration — `@primaryField` accepted by both rules.
  - Constraint-tag and metadata-slot branches exercised with `@`-prefixed and bare names.
  - Multi-extension merge (one annotation + one constraintTag across two extensions).
  - Negative cases: unrelated unknown tag still flagged when extensions exist; empty `extensions: []`; missing/`null` `settings.formspec`.
- [x] `pnpm --filter @formspec/eslint-plugin run test` — 319 passing.
- [x] `pnpm --filter @formspec/eslint-plugin run typecheck`, `lint` — clean.
- [x] `pnpm run format:check` — clean on modified files.

## Follow-ups (filed separately)

- Relocate `readExtensionTagNames` to `@formspec/analysis` so `@formspec/ts-plugin` and `@formspec/language-server` can reuse the same extension-tag flattening.
- Deduplicate the `context.settings.formspec.extensionRegistry` access currently duplicated between `utils/tag-metadata.ts` and `rules/type-compatibility/tag-type-check.ts`.

### DIFF
--- a/.changeset/fix-350-eslint-annotation-tags.md
+++ b/.changeset/fix-350-eslint-annotation-tags.md
@@ -1,0 +1,5 @@
+---
+"@formspec/eslint-plugin": patch
+---
+
+Fix `tag-recognition/no-unknown-tags` and `tag-recognition/tsdoc-comment-syntax` rejecting extension-registered annotation tags (e.g. `@primaryField`). Both rules now iterate `extension.annotations` in addition to `constraintTags` and `metadataSlots`.
--- a/packages/eslint-plugin/src/__tests__/rules/tag-recognition.test.ts
+++ b/packages/eslint-plugin/src/__tests__/rules/tag-recognition.test.ts
@@ -1,3 +1,13 @@
+/**
+ * Tests for tag-recognition rules: no-unknown-tags, require-tag-arguments,
+ * no-disabled-tags, no-markdown-formatting.
+ *
+ * Regression tests for issue #350: extension-registered annotation tags,
+ * constraint tags, and metadata slots supplied via `context.settings` were not
+ * recognized by `no-unknown-tags` and were incorrectly flagged as unknown.
+ *
+ * @see https://github.com/mike-north/formspec/issues/350
+ */
 import { RuleTester } from "@typescript-eslint/rule-tester";
 import * as vitest from "vitest";
 import { noUnknownTags } from "../../rules/tag-recognition/no-unknown-tags.js";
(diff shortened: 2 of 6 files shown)


mike-north:
## Review feedback addressed

Commit a2e4659 addresses all requested changes:

**`readExtensionTagNames` hardening (`tag-metadata.ts`)**
- Return type changed from `readonly string[]` to `ReadonlySet<string>`; `.sort()` removed
- Interface fields changed from `string` to `unknown` to model untrusted `context.settings`
- `typeof === "string"` guards added before calling `normalizeFormSpecTagName` on each tag/annotation name
- Non-object extension entries are silently skipped

**Call site updates**
- `no-unknown-tags.ts`: uses returned `ReadonlySet` directly (no `new Set(...)` wrapper)
- `tsdoc-comment-syntax.ts`: spreads with `[...readExtensionTagNames(settings)]` at the call site

**New tests**
- Empty `extension.annotations: []` → unknown tags still reported (no silent allow-listing)
- `annotationName: "minimum"` shadows a built-in → built-in handling unaffected (documented in test comment)

**Follow-up issue filed**
Moving `readExtensionTagNames` to `@formspec/analysis` to eliminate drift risk: https://github.com/mike-north/formspec/issues/353

Issue #309, opened 2026-04-19
Large enum types inlined at every usage site can exceed server-side string limits


mike-north:
### Summary

When a TypeScript string union type with many members (e.g., ISO 4217 currency codes with ~157 values) is used across multiple fields in a config type, FormSpec inlines the full enum at each usage site in the generated JSON Schema instead of placing it in `$defs` and using `$ref`. This causes the serialized schema string to exceed server-side string length limits (e.g., 5000 characters).

### Context

This was discovered during the Billing extensions breaking change incident (`#ir-cursor-inaugurate`). The billing team's 1P app upload failed with:

```
Error: Invalid string: {"$s..."}}}; must be at most 5000 characters
```

The `{"$s..."` prefix indicates the server received a raw JSON Schema string (starting with `{"$schema":"https://json-schema.org/draft/2020-12/schema",...}`) that exceeded the 5000-character limit.

### Reproduction

Given a TypeScript type like:

```typescript
type Currency = "AED" | "AFN" | "ALL" | /* ...~157 ISO 4217 codes... */ | "ZWL";

interface MonetaryAmount {
  amount: number;
  currency: Currency;
}

interface InvoiceConfig {
  subtotal: MonetaryAmount;
  total: MonetaryAmount;
  tax: MonetaryAmount;
  discount: MonetaryAmount;
  amountDue: MonetaryAmount;
}
```

**Expected behavior:** The generated JSON Schema should use `$defs` for `Currency` and `MonetaryAmount`, referencing them via `$ref`:

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "type": "object",
  "properties": {
    "subtotal": { "$ref": "#/$defs/MonetaryAmount" },
    "total": { "$ref": "#/$defs/MonetaryAmount" },
    "tax": { "$ref": "#/$defs/MonetaryAmount" },
    "discount": { "$ref": "#/$defs/MonetaryAmount" },
    "amount_due": { "$ref": "#/$defs/MonetaryAmount" }
  },
  "$defs": {
    "Currency": { "type": "string", "enum": ["AED", "AFN", ...] },
    "MonetaryAmount": {
      "type": "object",
      "properties": {
        "amount": { "type": "number" },
        "currency": { "$ref": "#/$defs/Currency" }
      }
    }
  }
}
```

This would keep the schema compact (~1700 chars).

[...]

PR #365, opened 2026-04-21
fix(build): use $ref sibling keywords instead of allOf (#364)

### LINKED ISSUE #364: allOf composition when field-level constraints are applied to $ref-based types breaks downstream renderers
### Summary

When a config property references a `$defs` type (e.g., `MonetaryAmount`) AND has field-level constraints applied to a nested property (e.g., `@exclusiveMinimum 0` on `amount`), FormSpec generates an `allOf` composition. Downstream renderers (Stripe dashboard config UI) don't support `allOf` and show a "no renderer found" error.

### Current behavior

TypeScript source:
```typescript
interface MinAmountConfig {
  /**
   * @displayName Minimum amount
   * @exclusiveMinimum 0
   */
  minimumAmount: MonetaryAmount;
}
```

Generated schema:
```json
{
  "minimum_amount": {
    "allOf": [
      { "$ref": "#/$defs/monetary_amount" },
      {
        "properties": {
          "amount": { "exclusiveMinimum": 0 }
        },
        "title": "Minimum amount"
      }
    ]
  }
}
```

### Expected behavior

One of:
1. **Inline the `$defs` entry** when constraints are applied, producing a flat object schema with the constraint merged in (loses deduplication but avoids `allOf`)
2. **Use `$ref` with sibling keywords** — JSON Schema 2020-12 allows keywords alongside `$ref`:
```json
{
  "minimum_amount": {
    "$ref": "#/$defs/monetary_amount",
    "properties": {
      "amount": { "exclusiveMinimum": 0 }
    },
    "title": "Minimum amount"
  }
}
```

### Context

- The `$defs` deduplication was introduced in alpha.55 (#309) to keep schema sizes under the 5000-char limit
- The `allOf` is semantically correct JSON Schema 2020-12 but breaks the Stripe dashboard config renderer
- Seiji's workaround: remove the `@exclusiveMinimum` annotation from the MonetaryAmount field

### Impact

Billing team's 1P app config with MonetaryAmount + `@exclusiveMinimum` renders as "no renderer found" in the dashboard.

Reported by: seiji, escalated by ericfrank

mike-north:
## Summary

- Fixes [#364](https://github.com/mike-north/formspec/issues/364): when a field referencing a `$defs` type also carries path-targeted constraints or annotations, FormSpec emitted an `allOf` wrapper that breaks downstream renderers (Stripe dashboard config UI).
- The fix switches `applyPathTargetedConstraints` to emit sibling keywords alongside `$ref`, which JSON Schema 2020-12 §10.2.1 explicitly permits (draft-07's "siblings next to `$ref` are ignored" rule was removed).
- `$defs` deduplication (introduced in alpha.55 via #309) is preserved — the referenced type still appears in `$defs` and is reused across call sites.

## Before / After

TypeScript source:
```typescript
interface MinAmountConfig {
  /**
   * @displayName Minimum amount
   * @exclusiveMinimum :value 0
   */
  minimumAmount: MonetaryAmount;
}
```

```diff
 {
   "minimumAmount": {
-    "allOf": [
-      { "$ref": "#/$defs/MonetaryAmount" },
-      {
-        "properties": { "value": { "exclusiveMinimum": 0 } },
-        "title": "Minimum amount"
-      }
-    ]
+    "$ref": "#/$defs/MonetaryAmount",
+    "properties": { "value": { "exclusiveMinimum": 0 } },
+    "title": "Minimum amount"
   }
 }
```

## Spec alignment

`docs/003-json-schema-vocabulary.md` §5.4 and §7.2 now declare sibling keywords alongside `$ref` as the canonical refinement shape. Cross-referenced examples in `docs/005-numeric-types.md` and `docs/e2e-test-matrix.md` (and the "putting it all together" example in 003 Appendix A) are updated to match.

## Scope note

Two remaining `allOf` emission sites in `applyPathTargetedConstraints` cover *inline-object* cases (missing-property overrides at [ir-generator.ts:517](https://github.com/mike-north/formspec/blob/fix/364-allof-composition-with-ref/packages/build/src/json-schema/ir-generator.ts#L517) and already-composed `allOf` schemas at line 523). Those are legitimately out of scope for #364 (which reports `$ref + allOf`), but share the same renderer-compatibility concern — tracked in [#382](https://github.com/mike-north/formspec/issues/382).

## Test plan

- [x] Primary regression suite [`packages/build/src/__tests__/ref-with-sibling-keywords.test.ts`](packages/build/src/__tests__/ref-with-sibling-keywords.test.ts) — 10 cases covering the primary bug, metadata combination, `$defs` deduplication, array items, multi-property overrides, annotation-only fields, deeply-nested path targets, and **top-level nullable `$ref`** (added in response to PR review — covers the `field: SomeRef | null` recursion at `ir-generator.ts:467-480`).
- [x] Narrowing-alias integration coverage in [`packages/build/src/__tests__/generate-schemas.test.ts`](packages/build/src/__tests__/generate-schemas.test.ts) — `type PositiveMonetaryAmount = MonetaryAmount` with `/** @minimum :amount 0 */` emits `$ref` + sibling `properties` at the field use site, and composes with a use-site `@maximum :amount N` override as further siblings on the same subschema.
- [x] Existing tests asserting the old `allOf` shape updated to the new sibling-keyword shape (4 files) — no assertion strength lost.
- [x] Every assertion pairs a positive check (sibling keyword present) with a negative check (`allOf` absent).
- [x] `pnpm run clean && pnpm install && pnpm run build && pnpm run lint && pnpm run test && pnpm run test:e2e` — all green against current `origin/main`.
- [x] `pnpm run api-extractor` — no public API change.

### DIFF
--- a/.changeset/fix-364-ref-with-sibling-keywords.md
+++ b/.changeset/fix-364-ref-with-sibling-keywords.md
@@ -1,13 +1,7 @@
 ---
-"@formspec/analysis": patch
 "@formspec/build": patch
 "@formspec/cli": patch
-"@formspec/config": patch
-"@formspec/dsl": patch
 "@formspec/eslint-plugin": patch
-"@formspec/language-server": patch
-"@formspec/runtime": patch
-"@formspec/ts-plugin": patch
 "formspec": patch
 ---
 
--- a/.changeset/fix-364-ref-with-sibling-keywords.md
+++ b/.changeset/fix-364-ref-with-sibling-keywords.md
@@ -1,0 +1,14 @@
+---
+"@formspec/analysis": patch
+"@formspec/build": patch
+"@formspec/cli": patch
+"@formspec/config": patch
+"@formspec/dsl": patch
+"@formspec/eslint-plugin": patch
+"@formspec/language-server": patch
+"@formspec/runtime": patch
+"@formspec/ts-plugin": patch
+"formspec": patch
+---
+
+Fix unnecessary `allOf` composition when field-level constraints or annotations are applied to `$ref`-based types. JSON Schema 2020-12 (§10.2.1) allows sibling keywords next to `$ref`, so the generator now emits `{ "$ref": "#/$defs/X", "properties": {...} }` directly instead of wrapping in `allOf`. This preserves `$defs` deduplication and produces output that downstream renderers can consume without needing to unwrap `allOf` as a workaround.
--- a/docs/003-json-schema-vocabulary.md
+++ b/docs/003-json-schema-vocabulary.md
@@ -429,26 +429,22 @@ All `$defs` entries are placed at the root schema level. Nested schemas do not h
 
 When an author applies subfield constraints (via `@minimum :value 0` on a `MonetaryAmount` field — see document 002 §4), the generator must decide whether to:
 
-**Option A: Inline refinement at the use site** — emit the `$ref` wrapped in an `allOf` that adds the constraints:
+**Option A: Sibling keywords alongside `$ref`** — emit the `$ref` with the refinement keywords as siblings of the reference:
 
 ```json
 {
-  "allOf": [
-    { "$ref": "#/$defs/MonetaryAmount" },
-    {
-      "properties": {
-        "value": { "minimum": 0 }
-      }
-    }
-  ]
+  "$ref": "#/$defs/MonetaryAmount",
+  "properties": {
+    "value": { "minimum": 0 }
+  }
 }
 ```
 
-**Option B: Derived `$defs` entry** — create a new named `$defs` entry (e.g., `MonetaryAmount_constrained_discount`) that `allOf`s the base type with the refinements.
+**Option B: Derived `$defs` entry** — create a new named `$defs` entry (e.g., `MonetaryAmount_constrained_discount`) that composes the base type with the refinements.
 
-**Decision: Option A (inline `allOf` at the use site) is the default.** Option B is reserved for future work when multiple fields share the same constraint profile and deduplication is beneficial. Option A is simpler, more readable, and correctly represents that the constraints are field-specific, not type-specific.
+**Decision: Option A (sibling keywords alongside `$ref`) is the default.** Option B is reserved for future work when multiple fields share the same constraint profile and deduplication is beneficial. Option A is simpler, more readable, and correctly represents that the constraints are field-specific, not type-specific.
 
-The `allOf` + `$ref` pattern is standard 2020-12 and is the idiomatic way to express "this type, but with additional constraints." Standards-compliant validators handle it correctly.
+Sibling keywords alongside `$ref` are standard JSON Schema 2020-12. Draft 2019-09 changed `$ref` so it no longer ignores adjacent keywords — the `$ref` and its siblings are evaluated together. Standards-compliant 2020-12 validators honor both halves. This is also the shape that downstream renderers expect: wrapping in `allOf` adds a layer of composition that many renderers do not unwrap, resulting in the constraints being silently dropped from the rendered view.
 
 ### 5.5 Circular Reference Handling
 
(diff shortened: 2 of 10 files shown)


mike-north:
Post-merge audit of this PR against current `main`. The core change in `ir-generator.ts` is clean and the test suite is solid, but the branch base is stale and the spec doc needs to come along. Blockers and a quick summary of what's already solid below.

## Blockers before merge

### 1. Rebase onto current `main` to restore shipped fixes

Branch forked from `ddfaca6d`; three fixes have since merged to `main` and are missing here:

- #359 — `coerceDefaultValue` for custom-typed `@defaultValue`
- #369 — heritage `@format` inheritance through class/interface `extends`
- #372 — NoTruncation regression coverage

As a result, `git diff origin/main..HEAD --name-status` shows this PR deleting:

- `.changeset/fix-358-default-value-custom-type-coercion.md`
- `.changeset/fix-format-inheritance-derived-types.md`
- `.changeset/strengthen-notruncation-regression-test.md`
- `packages/build/src/__tests__/default-value-custom-type-coercion.test.ts` (341 lines)
- `packages/build/src/__tests__/format-inheritance-derived-types.test.ts` (456 lines)

It also removes `CustomTypeRegistration.serializeDefault` from the `@alpha` public API surface (visible in both `packages/core/api-report/core.api.md` and `packages/build/api-report/*.api.md`). A clean rebase resolves all of this.

### 2. Update spec doc 003

`docs/003-json-schema-vocabulary.md` still declares the old `allOf + $ref` shape as canonical:

- §5.4 (lines 428–451) — "Option A: inline `allOf` at the use site" is the default decision.
- §7.2 (lines 539–556) — "`allOf` for Progressive Refinement" section describes the old shape as idiomatic.

Cross-references in other docs that reference §5.4 or carry stale `allOf + $ref` examples:

- `docs/005-numeric-types.md` lines 191, 197, 676, 684, 699, 788
- `docs/e2e-test-matrix.md` lines 26, 1011, 1014, 1157, 1169, 1190, 1206, 1221, 1272, 1274, 1782, 1795

Should land in this PR so the spec and the emitter don't disagree.

### 3. File (and link) a follow-up for the remaining `allOf` emissions

The PR body already scopes these out — but no tracking issue is linked, and the renderer compatibility motivation still applies to any consumer that hits them:

- `packages/build/src/json-schema/ir-generator.ts:517-519` — inline-object missing-property fallback (preserves `additionalProperties`).
- `packages/build/src/json-schema/ir-generator.ts:522-526` — pre-composed `allOf` schemas get further `allOf` appends.

### 4. Add a top-level nullable-ref regression test

`field: SomeRef | null` with a path-targeted constraint exercises the `nullableValueBranch` → `$ref` recursion at `packages/build/src/json-schema/ir-generator.ts:467-480`. Nested-nullable is covered at `ir-json-schema-generator.test.ts:2559-2654`; the top-level case is not.

## What's already solid

- The single-point change at `ir-generator.ts:492-497` is simpler than alternatives and explicitly documents the `properties`-overwrite invariant inline (better than prior attempts in this area).
- 8 new regression tests in `ref-with-sibling-keywords.test.ts` cover the main scenarios (basic `$ref` + numeric, multi-segment paths, array of refs, `$defs` dedup preservation, annotation-only, mixed keywords).
- Existing `allOf` assertions flip cleanly across `e2e/tests/tsdoc-path-target.test.ts`, `generate-schemas.test.ts`, `generate-schemas-config.test.ts`, `ir-json-schema-generator.test.ts` — no weakened intent.
- `findPathOverride` / `expectPathOverride` helper rewrites verified against all consumer call sites.
- Changeset correctly lists `@formspec/build`, `@formspec/cli`, `@formspec/eslint-plugin`, and `formspec` (per repo convention).
- Full suite green on an ad-hoc rebased worktree (835 `@formspec/build` + 270 e2e).
- `checkConstraintBroadening` (`packages/analysis/src/semantic-targets.ts:937`) is untouched — narrowing semantics preserved.
- Dialect is consistent: emitter declares `2020-12` and `@formspec/validator` defaults to `2020-12`, so `$ref` + sibling keywords evaluate as intended internally.

## Out of scope / defense-in-depth (not blockers)

- Stripe dashboard renderer smoke-test with the new shape (external verification, not something CI can check).
- Overlay vs. base keyword-conflict test (e.g., overlay `@format` on a nested property whose base already has `format`) — documents sibling-keyword evaluation semantics.
- Validator round-trip test for `$ref` + sibling `properties` against `@formspec/validator` — confirms 2020-12 validators honor the siblings end-to-end.

PR #389, opened 2026-04-21
fix(build): flatten remaining allOf emission sites to siblings (#382)

### LINKED ISSUE #364: allOf composition when field-level constraints are applied to $ref-based types breaks downstream renderers

### LINKED ISSUE #382: Emitter: replace remaining `allOf` emission sites with sibling keywords in ir-generator
Follow-up from #365 (fix #364). That PR replaced the primary `$ref`-path-target `allOf` composition with sibling keywords. Two `allOf` emission sites in [`packages/build/src/json-schema/ir-generator.ts`](packages/build/src/json-schema/ir-generator.ts) were intentionally scoped out and remain:

### Site 1 — inline-object missing-property fallback (`:517-519`)

```ts
// Do not introduce new properties directly; compose via allOf instead
// to preserve additionalProperties semantics on the base object.
return {
  allOf: [schema, { properties: missingOverrides }],
};
```

Triggered when a path-targeted override names a property that doesn't exist on an inline-object base schema. Today it wraps the base in `allOf` to keep `additionalProperties` semantics from leaking onto a synthesized new property.

### Site 2 — pre-composed `allOf` append (`:522-526`)

```ts
// allOf schema (already composed): add property overrides as another member.
if (schema.allOf) {
  schema.allOf = [...schema.allOf, { properties: propertyOverrides }];
  return schema;
}
```

Triggered when the base schema is already an `allOf` (e.g., from an intersection or earlier composition). Overrides are appended as another `allOf` member.

## Why this matters

The renderer-compatibility motivation in #364 still applies here: any downstream consumer that hits these code paths will receive the old `allOf`-wrapped shape, which the Stripe dashboard renderer (and others like it) do not unwrap — so the overrides are silently dropped from the rendered view.

## Acceptance criteria

- Site 1: emit the base object and the new property without `allOf`, preserving `additionalProperties` semantics. Concretely: spread the base `properties` with the missing overrides merged in, and carry over `additionalProperties`/`type`/etc. as siblings on a single flat object.
- Site 2: flatten a pre-composed `allOf` into sibling keywords when the composition is expressible as siblings (base `$ref` + sibling `properties`). Fall back to `allOf` only when the composition genuinely cannot be expressed as siblings under 2020-12 evaluation semantics.
- Regression coverage in `packages/build/src/__tests__/ref-with-sibling-keywords.test.ts` (or a neighbour) for both scenarios.
- Spec doc 003 §5.4 / §7.2 already declares siblings as canonical — confirm the new behaviour matches.

## Context

- #364 — original bug (allOf-wrapping breaks Stripe dashboard renderer)
- #365 — primary fix (use sibling keywords for `$ref` + path-target overrides)
- Related call sites and invariants are documented in the fix commit on #365 at `ir-generator.ts:482-497`.

mike-north:
## Summary

Follow-up to #365 (which fixed #364). Replaces the two remaining `allOf` emission sites in `packages/build/src/json-schema/ir-generator.ts` with flat sibling-keyword shapes, so downstream renderers that don't unwrap `allOf` (e.g. the Stripe dashboard renderer) see the path-targeted overrides.

### Site 1 — inline-object missing-property fallback

Previously wrapped the base in `allOf: [schema, { properties: missingOverrides }]`. Now merges missing overrides directly into `schema.properties` alongside any existing overrides. Under JSON Schema 2020-12, declaring a key in `properties` legitimizes it even under `additionalProperties: false`, so the old composition was unnecessary.

### Site 2 — pre-composed `allOf` append

Previously appended `{ properties: propertyOverrides }` as another `allOf` member. Now flattens to sibling keywords when safe — via a new helper `tryFlattenAllOfToSiblings` — and falls back to the old `allOf` append only when the composition genuinely cannot be expressed as siblings (multiple members, or a key collision that would silently overwrite a contribution).

## Acceptance criteria (from #382)

- [x] Site 1: single flat object, no `allOf`, `additionalProperties` semantics preserved
- [x] Site 2: flatten to siblings when expressible; retain `allOf` only for genuine non-sibling composition
- [x] Regression coverage in `ref-with-sibling-keywords.test.ts` (6 new tests covering both sites, both flatten and fallback paths)
- [x] Matches spec doc 003 §5.4 / §7.2 (and JSON Schema 2020-12 §10.2.1)

## Test plan

- [x] `pnpm run build` — clean
- [x] `pnpm run typecheck` — clean
- [x] `pnpm run lint` — clean
- [x] `pnpm --filter @formspec/build run test` — 867 passed, 4 skipped (pre-existing)
- [x] New regression suite in `ref-with-sibling-keywords.test.ts` passes and fails against pre-fix code (every key assertion is `expect(allOf).toBeUndefined()`)

## Reviewer notes

Reviewed in parallel by typescript-expert, simplicity-expert, testing-expert, code-quality-expert, and system-architect. Nits applied in a follow-up commit:
- Use `Set` for the override-key collision check (O(1) vs O(n))
- JSDoc note explaining the single-member restriction reflects current emitter/external-hook usage rather than a 2020-12 semantic limit
- Annotate the `noUncheckedIndexedAccess`-required undefined guard

The architect flagged a medium-severity "missed unification opportunity": the three sibling-keyword sites (#365's `$ref`, Site 1, Site 2) could share a single `composeAsSiblings` primitive. Deferred as a follow-up — the current shapes are different enough that forcing a shared helper would hurt readability today.

Closes #382.

### DIFF
--- a/packages/build/src/__tests__/ref-with-sibling-keywords.test.ts
+++ b/packages/build/src/__tests__/ref-with-sibling-keywords.test.ts
@@ -892,9 +892,7 @@ describe("remaining allOf emission sites flattened to siblings — issue #382", 
         ]);
 
         const schema = generateJsonSchemaFromIR(ir);
-        const payload = schema.properties?.["payload"] as
-          | Record<string, unknown>
-          | undefined;
+        const payload = schema.properties?.["payload"] as Record<string, unknown> | undefined;
         const props = payload?.["properties"] as Record<string, unknown> | undefined;
 
         // Object.prototype.constructor still points at Object — we did not
@@ -938,9 +936,7 @@ describe("remaining allOf emission sites flattened to siblings — issue #382", 
         ]);
 
         const schema = generateJsonSchemaFromIR(ir);
-        const payload = schema.properties?.["payload"] as
-          | Record<string, unknown>
-          | undefined;
+        const payload = schema.properties?.["payload"] as Record<string, unknown> | undefined;
         const props = payload?.["properties"] as Record<string, unknown> | undefined;
 
         // The override landed as a real own property — Object.hasOwn returns
@@ -949,14 +945,12 @@ describe("remaining allOf emission sites flattened to siblings — issue #382", 
         // `obj["__proto__"] = value` would have replaced [[Prototype]]
         // instead, leaving no own property.
         expect(Object.hasOwn(props ?? {}, "__proto__")).toBe(true);
-        expect(
-          Object.getOwnPropertyDescriptor(props, "__proto__")?.value
-        ).toEqual({ minLength: 1 });
+        expect(Object.getOwnPropertyDescriptor(props, "__proto__")?.value).toEqual({
+          minLength: 1,
+        });
         // Object.prototype is untouched — the global prototype chain is not
         // mutated as a side-effect of emission.
-        expect(Object.getOwnPropertyNames(Object.prototype).sort()).toEqual(
-          protoSnapshot
-        );
+        expect(Object.getOwnPropertyNames(Object.prototype).sort()).toEqual(protoSnapshot);
       });
     });
   });
--- a/packages/build/src/__tests__/ref-with-sibling-keywords.test.ts
+++ b/packages/build/src/__tests__/ref-with-sibling-keywords.test.ts
@@ -26,10 +26,12 @@ import type {
   FieldNode,
   ConstraintNode,
   AnnotationNode,
+  CustomTypeNode,
   Provenance,
 } from "@formspec/core/internals";
-import { IR_VERSION } from "@formspec/core/internals";
+import { defineCustomType, defineExtension, IR_VERSION } from "@formspec/core/internals";
 import { generateJsonSchemaFromIR } from "../json-schema/ir-generator.js";
+import { createExtensionRegistry } from "../extensions/index.js";
 
 // =============================================================================
 // TEST HELPERS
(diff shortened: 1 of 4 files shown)


packages/build/src/json-schema/ir-generator.ts
```diff
@@ -496,32 +496,50 @@ function applyPathTargetedConstraints(
...
-      if (schema.properties[target]) {
-        mergeSchemaOverride(schema.properties[target], overrideSchema);
-      } else {
-        // Do not introduce new properties directly; compose via allOf instead
-        // to preserve additionalProperties semantics on the base object.
-        missingOverrides[target] = overrideSchema;
+      // Own-property check + defineProperty guard against prototype-pollution
+      // vectors (e.g. `__proto__`, `constructor`) if a path-targeted override
+      // ever names one of those keys. Using `in` or `[target]` assignment
+      // would traverse / mutate Object.prototype.
```
mike-north: The `in` operator is a read-only prototype-chain lookup — it doesn't mutate `Object.prototype`. Only plain-assignment (`obj[target] = value` with `target === "__proto__"`) triggers the `__proto__` setter and mutates the object's `[[Prototype]]`. The real hazard with `in` here is a false-positive membership match via inheritance (e.g. `"constructor" in {}` returns `true`), which would cause the existing-property branch to fire and merge into the inherited `constructor` value. Suggested wording clarifies both:

```suggestion
      // Own-property lookup + defineProperty guard against prototype-pollution
      // vectors when a path target names a key like `__proto__`:
      //   - Plain `obj[target] = value` assignment with target === "__proto__"
      //     invokes the Object.prototype `__proto__` setter, replacing the
      //     object's [[Prototype]] instead of adding an own property.
      //   - `Object.defineProperty` bypasses the setter and writes an own
      //     data property, so the override lands where we expect.
      // `Object.hasOwn` (not `in`) rejects inherited members like
      // `constructor`, avoiding a mis-merge into Object.prototype.constructor.```

File: packages/build/src/json-schema/ir-generator.ts [hunk 2]
```diff
@@ -496,32 +496,50 @@ function applyPathTargetedConstraints(
...
+      if (Object.hasOwn(schema.properties, target)) {
```
mike-north, line 519: The `Object.hasOwn` + `Object.defineProperty` guard here closes the `constructor` vector, but the `__proto__` vector is still open upstream: `buildPropertyOverrides` (outside this PR's diff, around line 947 in the same file) constructs `propertyOverrides` via `overrides[resolveSerializedPropertyName(target, ...)] = ...`. When the segment is `"__proto__"`, that bracket assignment invokes the `Object.prototype.__proto__` setter and replaces `overrides`'s prototype instead of creating an own property. By the time control reaches this loop, `Object.entries(propertyOverrides)` returns `[]` for the `__proto__` case and the guard never runs.

Demonstration:
```js
const o = {}; o["__proto__"] = { title: "x" };
Object.entries(o);             // []
Object.hasOwn(o, "__proto__");  // false
```

To fully close the vector, either construct `overrides` with `Object.create(null)` in `buildPropertyOverrides`, or use `Object.defineProperty` at that upstream assignment site. The PR description and commit message for d5b94f01 frame the hardening as closing both `__proto__` and `constructor` — today only `constructor` is actually closed. Worth either tightening the scope in this PR or documenting `__proto__`-named path segments as silently dropped.

packages/build/src/__tests__/ref-with-sibling-keywords.test.ts
```diff
@@ -645,3 +647,390 @@ describe("$ref with sibling keywords — issue #364", () => {
...
+//           composition is expressible as siblings under JSON Schema 2020-12
+//           (§10.2.1). `allOf` is retained only for genuine non-sibling cases.
+//
+// @see https://github.com/mike-north/formspec/issues/382
+
+describe("remaining allOf emission sites flattened to siblings — issue #382", () => {
+  // ---------------------------------------------------------------------------
+  // Site 1: inline-object missing-property fallback
+  // ---------------------------------------------------------------------------
+  describe("Site 1: inline-object with path target naming a property that does not exist", () => {
```
mike-north: The prototype-pollution hardening at `ir-generator.ts:510-526` (`Object.hasOwn` + `Object.defineProperty`, added in commit d5b94f01 specifically to defend against `__proto__` and `constructor` path targets) has no regression test. The strings `__proto__` and `constructor` do not appear anywhere in this file or in `ir-json-schema-generator.test.ts`. Without a test, a future refactor that reverts to `schema.properties[target] = overrideSchema` would silently reopen the vector.

Suggested addition inside the Site 1 describe block (matches existing fixture style):

```typescript
it("declares a path target named `constructor` as an own property without touching Object.prototype.constructor", () => {
  // Regression guard for the prototype-pollution hardening at ir-generator.ts:510-526.
  // Without Object.hasOwn + defineProperty, `\"constructor\" in schema.properties` would
  // match the inherited Object.prototype.constructor and the override would silently
  // merge into the wrong object. @see https://github.com/mike-north/formspec/issues/382
  // ... build IR with path: { segments: [\"constructor\"] } ...
  const schema = generateJsonSchemaFromIR(ir);
  expect(Object.prototype.constructor).toBe(Object);
  expect(Object.hasOwn(props, "constructor")).toBe(true);
  expect(props["constructor"]).toEqual({ minLength: 1 });
});

it("handles a path target named `__proto__` without mutating the prototype chain", () => {
  // A plain `obj[\"__proto__\"] = v` replaces the object's prototype instead of
  // creating an own property; defineProperty is the safe form.
  // NOTE: upstream buildPropertyOverrides currently drops __proto__-named segments
  // before they reach Site 1 — this test should assert whichever behavior is the
  // intended contract (see the separate comment about buildPropertyOverrides).
});
```

Five of eight reviewers independently flagged this gap.

packages/build/src/__tests__/ref-with-sibling-keywords.test.ts
```diff
@@ -645,3 +647,390 @@ describe("$ref with sibling keywords — issue #364", () => {
...
+  //
+  // Pre-composed `allOf` schemas reach `applyPathTargetedConstraints` when a
+  // custom-type extension returns an `allOf` schema from `toJsonSchema`. The
+  // fix flattens to siblings when expressible under JSON Schema 2020-12 (i.e.,
+  // the allOf has exactly one member that is a `$ref`-shaped schema whose
+  // keys do not conflict with the overrides), and retains `allOf` only when
+  // the composition cannot be expressed as siblings.
+  //
+  // @see https://github.com/mike-north/formspec/issues/382 Site 2
+  describe("Site 2: pre-composed allOf base", () => {
```
mike-north: `tryFlattenAllOfToSiblings` checks three pairwise overlaps: `memberKeys ∩ outerKeys` (line 1040), `memberKeys ∩ overrideKeys` (line 1040), and `outerKeys ∩ overrideKeys` (lines 1044-1048). The test added further down this block covers the `memberKeys ∩ overrideKeys` rejection. The other two rejection branches (outer↔member, outer↔override) have no direct coverage.

These branches are currently unreachable from in-tree callers (emitter-produced outer schemas carry no keywords besides `allOf`), but extension-supplied `toJsonSchema` hooks can return outer schemas with top-level keywords — exactly the shape this fallback exists to handle. A silent regression in either branch would not be caught.

Suggested shape (one test per branch):
```typescript
it("retains allOf when an outer sibling keyword would collide with a member key", () => {
  // Custom type returns { type: \"object\", allOf: [{ type: \"string\" }] }
  // `type` collides with the outer `type` — flattening would silently overwrite.
  // Assert: emitted schema retains allOf under this input.
});

it("retains allOf when an outer sibling keyword would collide with the override key", () => {
  // Custom type returns { properties: {...}, allOf: [...] }
  // Path-targeted override would contribute `properties` — flattening would overwrite outer.
});
```

packages/build/src/json-schema/ir-generator.ts
```diff
@@ -971,6 +989,71 @@ function buildPathOverrideSchema(
...
+ *
+ * The single-member restriction reflects what the emitter produces today
+ * (external `toJsonSchema` hooks may legitimately return single-member
+ * `allOf` for wrapping). Generalising to N-member flattening is possible
+ * via pairwise key-disjointness but is not needed by any current caller.
+ *
+ * Mirrors the `$ref`-sibling fix at `ir-generator.ts:492-497` (issue #364).
+ *
+ * @see https://github.com/mike-north/formspec/issues/382 Site 2
+ * @see https://json-schema.org/draft/2020-12/json-schema-core — §10.2.1 sibling keywords
```
mike-north, line 1037: The single-member restriction is presented as "what the emitter produces today," but the Site 2 path is specifically the one reached when user-supplied `toJsonSchema` hooks return pre-composed `allOf` — which is not the in-tree emitter. The framing may mislead a future reader who adds a caller. Consider clarifying that the restriction is a conservative domain-scope choice (not a 2020-12 semantic limit) and that multi-member `allOf` from extension hooks is where pairwise-disjoint N-member flattening would be added if ever needed:

```suggestion
 * The single-member restriction is a conservative scope choice, not a
 * JSON Schema 2020-12 semantic constraint. In-tree emission paths only
 * ever produce single-member `allOf` wrappers. Multi-member `allOf`
 * reaches this helper only from user-supplied `toJsonSchema` hooks,
 * where the multiple members typically represent intentional composition
 * that should not be silently flattened. Pairwise-disjoint N-member
 * flattening is semantically valid but is not performed today because
 * no current producer requires it.```

packages/build/src/__tests__/ref-with-sibling-keywords.test.ts
mike-north: Append the issue number to the describe label so failing CI output immediately traces back to the tracked issue (matches the file-level precedent `"$ref with sibling keywords — issue #364"` at line 102):

```suggestion
  describe("Site 1: inline-object with path target naming a property that does not exist (#382)", () => {```

packages/build/src/__tests__/ref-with-sibling-keywords.test.ts
mike-north: Same housekeeping as Site 1 — append the issue number to the describe label:

```suggestion
  describe("Site 2: pre-composed allOf base (#382)", () => {```

mike-north: The three Site 2 tests each inline the full `FormIR` wrapper (`kind: "form-ir"`, `irVersion`, `elements`, `typeRegistry`, `provenance`) instead of using the `makeIR` helper defined at line 55 and used throughout the #364 block above. Adopting `makeIR` here would remove roughly 70 lines of boilerplate and make the actual variation between the three tests (the custom-type registry payload) the visible story. Per the project's TypeScript-testing rule ("extract reusable test helpers when setup/data is repeated"), this is the appropriate pattern.

Happy to defer to a follow-up if you prefer to keep this PR scoped to the emitter change.

Issue #424, opened 2026-04-26
TypeScript 6.x compatibility


mike-north:
## Outcome

Workspace builds, type-checks, and tests cleanly under TypeScript 6.x with no rows in the multi-version CI matrix marked `experimental` for the 6.x line. Once landed, the published `typescript` peer-dep range can be widened from `^5.9.3` to `>=5.9 <7`, and consumers can adopt FormSpec on TS 6 without warnings.

This epic explicitly excludes TypeScript 7.x (the Go rewrite, separate API surface) — that's a separate, deliberate migration whenever it's ready.

## Empirical state under TS 6.0.3

Verified locally on the current `main` (and on the `claude/practical-merkle-0428de` branch behind PR #412):

| Stage | Result |
|---|---|
| `pnpm run build` (all 13 packages) | passes |
| `pnpm run typecheck` (all 13 packages) | passes |
| `pnpm run test` — 12 of 13 packages | passes |
| `pnpm run test` — `@formspec/eslint-plugin` | **29 failures across 3 test files** |

The substantive TS 6 surface is concentrated in **one file** (`packages/eslint-plugin/src/utils/type-utils.ts`) with **one root cause** (hardcoded `ts.TypeFlags` numeric literals that no longer match TS 6's renumbered enum).

### TypeFlags enum churn between TS 5.x and TS 6.x

| Flag | TS 5.x | TS 6.x |
|---|---|---|
| `Null` | 65536 | 8 |
| `Undefined` | 32768 | 4 |
| `String` | 4 | 32 |
| `Number` | 8 | 64 |
| `Boolean` | 16 | 256 |
| `StringLiteral` | 128 | 1024 |
| `NumberLiteral` | 256 | 2048 |
| `BooleanLiteral` | 512 | 8192 |
| `BigInt` | 64 | 128 |
| `BigIntLiteral` | 2048 | 4096 |
| `Object` | 524288 | (changed) |

In TS 6, `type.flags & 65536` no longer detects `Null` — it picks up an unrelated flag. The eslint rules cascade from broken type-classification helpers and report wrong `messageId`s (e.g. `nonStringLikeTargetField` instead of `nullableTargetField`).

I verified the fix in a scratch worktree: replacing the hardcoded numbers with `ts.TypeFlags.X` enum references makes all 319 eslint-plugin tests pass under TS 6.0.3 and the full workspace test suite (1000+ tests) goes green.

## Sub-issues

[...]

PR #431, opened 2026-04-26
Support TypeScript 6.x

### LINKED ISSUE #425: Replace hardcoded ts.TypeFlags numbers in @formspec/eslint-plugin with enum references
Parent: #424

## Problem

[`packages/eslint-plugin/src/utils/type-utils.ts`](https://github.com/mike-north/formspec/blob/main/packages/eslint-plugin/src/utils/type-utils.ts) imports `typescript` as a type-only import and uses hardcoded numeric literals for `ts.TypeFlags` checks:

```ts
import type ts from "typescript";

if (type.flags & 4 /* ts.TypeFlags.String */) { ... }
if (type.flags & 65536 /* ts.TypeFlags.Null */) { ... }
if (type.flags & 32768 /* ts.TypeFlags.Undefined */) { ... }
// ...12 occurrences total
```

TypeScript 6.x renumbered the entire `TypeFlags` enum (table in #424). Under TS 6, `type.flags & 65536` no longer detects `Null` — it matches an unrelated flag. This cascades through `isStringType`, `isNumberType`, `isBooleanType`, `isNullableType`, and `getFieldTypeCategory`, producing wrong-`messageId` failures in:

- `tests/factories/create-constraint-rule.test.ts`
- `tests/rules/constraint-type-mismatch.test.ts` (`tag-type-check`)
- `tests/rules/constraint-validation-extra.test.ts` (`valid-discriminator`)

Total: 29 failing tests across 3 files when run under TS 6.0.3.

## Fix

1. Convert `import type ts from "typescript"` to `import * as ts from "typescript"` in `type-utils.ts`. `typescript` is already a peer-dep of `@formspec/eslint-plugin`, so no new runtime dependency.
2. Replace each `type.flags & <number>` literal with `type.flags & ts.TypeFlags.X`.

## Acceptance

- [ ] All 319 `@formspec/eslint-plugin` tests pass under TS 5.9.3 (current default).
- [ ] All 319 `@formspec/eslint-plugin` tests pass under TS 6.0.3.
- [ ] Full workspace `pnpm run test` passes under TS 6.0.3.
- [ ] PR #412's `Test (TypeScript latest (6.0.3))` matrix row goes green.

## Verified fix

I patched this locally in a scratch worktree and confirmed all three TS 6.x matrix rows go green and all 319 eslint-plugin tests pass — see commentary on #424. Reverted before pushing so this issue can capture the work cleanly.

### LINKED ISSUE #426: Bump typescript peer-dep range to support TS 6.x
Parent: #424

## Problem

The published peer-dep range in `@formspec/analysis`, `@formspec/build`, `@formspec/eslint-plugin`, `@formspec/ts-plugin` is `^5.9.3`, which excludes TypeScript 6.x. Once the workspace is functionally TS 6 compatible (sub-issue #425), the peer-dep claim should be widened so consumers on TS 6 don't see peer-dep warnings.

## Fix

Update each affected `package.json` `peerDependencies` block:

\`\`\`diff
- "typescript": "^5.9.3"
+ "typescript": ">=5.9.3 <7"
\`\`\`

Affected packages:

- `@formspec/analysis`
- `@formspec/build`
- `@formspec/eslint-plugin`
- `@formspec/ts-plugin`

(The `<7` upper bound is deliberate — TS 7.x is the Go rewrite with a different API surface. Tracked separately, not in scope of this epic.)

## Acceptance

- [ ] All four `package.json` peer-dep ranges updated.
- [ ] Changeset added at `minor` level (peer-dep widening is a public API change).
- [ ] No peer-dep warnings when installing into a project on TS 6.x.

## Dependencies

Should land **after or with** #425 — widening the peer-dep claim before the code actually works under TS 6 would be a regression for consumers.

### LINKED ISSUE #427: Promote TS 6.x matrix rows from experimental to required in CI
Parent: #424

## Problem

PR #412 introduced a multi-TypeScript-version CI matrix. Rows whose major doesn't match the workspace's declared TypeScript major are marked `experimental: true` (i.e. `continue-on-error: true`), so they don't block merges while real incompatibilities exist. This is correct for surfacing TS 6 issues today, but once #425 + #426 land, 6.x should block on regression.

## Fix

Two equivalent ways to flip this — pick one:

**Option A (preferred):** Bump the workspace's root `devDependencies.typescript` from `^5.9.3` to `^6.0.0`. The matrix compute logic in `.github/workflows/ci.yml` derives `experimental` from a major-version mismatch with that range, so 6.x rows auto-promote to required and a `5.x` row auto-enables for backwards-compat coverage.

**Option B:** Edit `.github/workflows/ci.yml`'s `typescript-matrix` job to remove the major-mismatch check, making `latest` (and any post-release stable) always required.

## Acceptance

- [ ] `Test (TypeScript latest (6.x))` is required for merge (no `continue-on-error`).
- [ ] `Test (TypeScript 5.x)` exists as a regression-coverage row.
- [ ] `Test (TypeScript beta)` and `Test (TypeScript 6.x nightly)` remain experimental — pre-releases are expected to fail occasionally.

## Dependencies

Must land **after** #425 and #426.

mike-north:
## Summary

Closes #425
Closes #426
Closes #427

Adopts TypeScript 6.x as a supported major across the workspace, addressing the three required items in the **TypeScript 6.x compatibility** epic ([#424](https://github.com/mike-north/formspec/issues/424)). Verified locally: build, typecheck, lint, and the full test suite (~2900 tests) all pass under both TS 5.9.3 and TS 6.0.3.

### #425 — Fix `ts.TypeFlags` cross-version drift in `@formspec/eslint-plugin`

`packages/eslint-plugin/src/utils/type-utils.ts` previously checked `ts.Type` kinds via hardcoded `TypeFlags` numeric literals. TS 6.x renumbered the entire enum, so `type.flags & 65536` no longer detected `Null`, and the broken type-classification helpers cascaded into 29 wrong-`messageId` failures across three rule test files.

Fix: switch the `typescript` import from type-only to value (it's already a peer-dep) and replace every hardcoded literal with its `ts.TypeFlags.X` enum reference. Behavior under TS 5.x is unchanged.

The file now leads with a thorough comment block — flag churn table, rule for future contributors, what to do if a hardcoded number is ever spotted again — because this is exactly the class of subtle, version-dependent bug that future contributors would have no context for.

### #426 — Widen `typescript` peer-dep range to `>=5.9.3 <7`

Updated in `@formspec/analysis`, `@formspec/build`, `@formspec/eslint-plugin`, `@formspec/ts-plugin`. The `<7` upper bound is deliberate — TypeScript 7.x is the Go rewrite with a substantively different API surface, and that migration will be its own deliberate effort.

`@formspec/cli`'s `dependencies.typescript` is also bumped to `^6.0.0` to match what the workspace actually develops and ships against.

### #427 — Promote 6.x matrix rows from experimental to required

Bumped the workspace's pinned `devDependencies.typescript` from `^5.9.3` to `^6.0.0`. The CI matrix derives `experimental` from a major-version mismatch with the workspace's declared TS major (per [.github/workflows/ci.yml](.github/workflows/ci.yml)), so this flip auto-promotes 6.x rows to required and auto-enables a 5.x backwards-compat row.

### Plumbing

- Added a `versionGroup` to `.syncpackrc.json` so the wider peer-dep range (`>=5.9.3 <7`) doesn't trip the default-version-group check against the workspace pin (`^6.0.0`).
- Aligned `e2e/package.json`'s devDep TS to `^6.0.0`.
- Added a "Supported TypeScript versions" section to `CLAUDE.md` summarizing the policy and pointing future contributors to the `type-utils.ts` notes when they're poking at the compiler API.

## Verification

| Stage | TS 5.9.3 | TS 6.0.3 |
|---|---|---|
| `pnpm run build` | passes | passes |
| `pnpm run typecheck` | passes | passes |
| `pnpm run lint` | passes | passes |
| `pnpm run test` | passes | passes |

## Test plan

- [ ] CI's `Test (TypeScript latest (6.x))` matrix row goes green (required after this lands)
- [ ] CI's `Test (TypeScript 5.x)` row exists and passes (now experimental, regression coverage)
- [ ] CI's `beta` and `6.x nightly` rows remain experimental
- [ ] No new peer-dep warnings under either TS major

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -138,11 +138,13 @@ pnpm run release            # Build and publish (CI usually does this)
 
 ## Supported TypeScript versions
 
-FormSpec packages with a `typescript` peer-dep declare `>=5.9.3 <7`. The CI matrix in `.github/workflows/ci.yml` exercises:
+FormSpec packages with a `typescript` peer-dep declare `>=5.7.3 <7`. The CI matrix in `.github/workflows/ci.yml` exercises:
 
 - the workspace's pinned TS major (currently 6.x — required, blocking),
-- the previous stable major (5.x — required, regression coverage),
+- the previous stable major (5.x — experimental, non-blocking regression coverage),
 - the `beta` and nightly dist-tags (experimental, non-blocking).
+
+The matrix derives the `experimental` flag from a major-version mismatch with the workspace's pinned TS major, so non-default majors are informational rather than gating. Once a TS major has had time to soak through real usage, we promote it to required by bumping the workspace's `devDependencies.typescript`.
 
 TypeScript 7.x is the Go rewrite with a substantively different API surface. We do not test against it via dist-tag drift; it will be a deliberate, separate migration.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -135,3 +135,21 @@ pnpm run release            # Build and publish (CI usually does this)
 - TSDoc constraint tags (`/** @minimum 0 @maximum 100 */`) are extracted via static AST analysis
 - Do not use `@description`; use summary text before block tags and `@remarks` for programmatic notes
 - API Extractor manages public API surface for library packages — commit `api-report/` files
+
+## Supported TypeScript versions
+
+FormSpec packages with a `typescript` peer-dep declare `>=5.9.3 <7`. The CI matrix in `.github/workflows/ci.yml` exercises:
+
+- the workspace's pinned TS major (currently 6.x — required, blocking),
+- the previous stable major (5.x — required, regression coverage),
+- the `beta` and nightly dist-tags (experimental, non-blocking).
+
+TypeScript 7.x is the Go rewrite with a substantively different API surface. We do not test against it via dist-tag drift; it will be a deliberate, separate migration.
+
+### TypeScript compiler API quirks across majors
+
+When code reaches into TypeScript's compiler API (`ts.Type`, `ts.Symbol`, `ts.TypeChecker`, etc.), be aware that some internals are not part of the public API contract and shift between majors. The most common pitfall:
+
+- **`ts.TypeFlags` numeric values were renumbered between TS 5.x and TS 6.x.** Always reference flags by enum member (`type.flags & ts.TypeFlags.Null`), never by hardcoded number (`type.flags & 65536`). See [`packages/eslint-plugin/src/utils/type-utils.ts`](packages/eslint-plugin/src/utils/type-utils.ts) for a full table and rationale.
+
+When in doubt, `import * as ts from "typescript"` (rather than `import type`) so enum references resolve at runtime against the host's installed TypeScript.
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -31,7 +31,7 @@
     "eslint": "^9.39.2",
     "stripe": "^22.0.2",
     "tsx": "^4.21.0",
-    "typescript": "^5.9.3",
+    "typescript": "^6.0.0",
     "vitest": "^3.2.4",
     "vscode-languageserver-textdocument": "^1.0.12"
   },
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -31,7 +31,7 @@
     "eslint": "^9.39.2",
     "stripe": "^22.0.2",
     "tsx": "^4.21.0",
-    "typescript": "^6.0.0",
+    "typescript": ">=5.7.3 <7",
     "vitest": "^3.2.4",
     "vscode-languageserver-textdocument": "^1.0.12"
   },
--- a/.changeset/cli-bump-typescript-dep-to-6.md
+++ b/.changeset/cli-bump-typescript-dep-to-6.md
@@ -1,0 +1,5 @@
+---
+"@formspec/cli": patch
+---
+
+Bump the bundled `typescript` runtime dependency from `^5.9.3` to `^6.0.0`. The CLI now performs schema generation against TypeScript 6.x internally. Source code that compiles cleanly under TS 5.9 also compiles under TS 6.0, so this is functionally invisible to consumers.
--- a/.changeset/eslint-plugin-typeflags-enum-refs.md
+++ b/.changeset/eslint-plugin-typeflags-enum-refs.md
@@ -1,0 +1,5 @@
+---
+"@formspec/eslint-plugin": patch
+---
+
+Fix `@formspec/eslint-plugin` rules under TypeScript 6.x by replacing hardcoded `ts.TypeFlags` numeric literals in the type-classification helpers with `ts.TypeFlags.X` enum references. TS 6 renumbered the entire `TypeFlags` enum, which caused `isStringType`, `isNumberType`, `isBooleanType`, `isNullableType`, and `getFieldTypeCategory` to produce wrong results (e.g. reporting `nonStringLikeTargetField` instead of `nullableTargetField` for `string | null`). Behavior under TS 5.x is unchanged.
--- a/.changeset/typescript-peer-dep-widen-to-6.md
+++ b/.changeset/typescript-peer-dep-widen-to-6.md
@@ -7,6 +7,6 @@
 "formspec": minor
 ---
 
-Widen the `typescript` peer-dep range from `^5.9.3` to `>=5.9.3 <7`. FormSpec now officially supports TypeScript 6.x in addition to 5.9+. The `<7` upper bound is deliberate — TypeScript 7.x is the Go rewrite with a substantively different API surface, and that migration will be handled separately.
+Widen the `typescript` peer-dep range from `^5.9.3` to `>=5.7.3 <7`. FormSpec now officially supports TypeScript 5.7 through 6.x. The `<7` upper bound is deliberate — TypeScript 7.x is the Go rewrite with a substantively different API surface, and that migration will be handled separately. The 5.7 floor reflects the lowest version where the workspace's full toolchain (build, typecheck, test, lint) passes end-to-end; build/test alone work down to 5.5, but `@typescript-eslint/parser` 8.x's project-service mode misbehaves below 5.7.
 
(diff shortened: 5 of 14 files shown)


e2e/package.json
```diff
@@ -31,7 +31,7 @@
     "eslint": "^9.39.2",
     "stripe": "^22.0.2",
     "tsx": "^4.21.0",
-    "typescript": "^5.9.3",
+    "typescript": "^6.0.0",
```
mike-north: Let's be very deliberate about whether we want to state a dependency on TypeScript 6 or whether we want to run compatibility tests against TypeScript 6.

CLAUDE.md
```diff
@@ -135,3 +135,21 @@ pnpm run release            # Build and publish (CI usually does this)
 - TSDoc constraint tags (`/** @minimum 0 @maximum 100 */`) are extracted via static AST analysis
 - Do not use `@description`; use summary text before block tags and `@remarks` for programmatic notes
 - API Extractor manages public API surface for library packages — commit `api-report/` files
+
+## Supported TypeScript versions
+
+FormSpec packages with a `typescript` peer-dep declare `>=5.9.3 <7`. The CI matrix in `.github/workflows/ci.yml` exercises:
```
mike-north: Can we see if we can push the oldest edge of this version range back to TypeScript 5.8 (or even earlier)?

CLAUDE.md
```diff
@@ -135,3 +135,21 @@ pnpm run release            # Build and publish (CI usually does this)
...
+
+- the workspace's pinned TS major (currently 6.x — required, blocking),
+- the previous stable major (5.x — required, regression coverage),
+- the `beta` and nightly dist-tags (experimental, non-blocking).
+
+TypeScript 7.x is the Go rewrite with a substantively different API surface. We do not test against it via dist-tag drift; it will be a deliberate, separate migration.
```
mike-north, line 149: TypeScript 7 is out in beta. Apparently there's a ts 6 compat package -- we should explore whether it's possible to get this testing as well https://devblogs.microsoft.com/typescript/announcing-typescript-7-0-beta/

e2e/package.json
mike-north: Addressed in 1293a5f5. Reverted `e2e/package.json` to `>=5.7.3 <7` (the supported peer-dep range, after also lowering the floor — see the other thread). The intent for `e2e` is to track the same range we declare for consumers so the matrix can override to any supported TS version freely; declaring a hard `^6.0.0` would have implied "e2e requires TS 6," which contradicts that goal. Added a syncpack `versionGroup` so this wider range does not trip the default-version-group check.

CLAUDE.md
mike-north: Investigated empirically by overriding `pnpm.overrides.typescript` and running the full pipeline (build + typecheck + test + lint) at each minor going back. Findings:

| TS version | build | typecheck | test | lint |
|---|---|---|---|---|
| 6.0.3 | OK | OK | OK | OK |
| 5.9.3 | OK | OK | OK | OK |
| 5.8.3 | OK | OK | OK | OK |
| 5.7.3 | OK | OK | OK | OK |
| 5.6.3 | OK | OK | OK | **fail** |
| 5.5.4 | OK | OK | OK | **fail** |
| 5.4.5 | **fail** (control-flow narrowing in `class-analyzer.ts`, `ir-generator.ts`) | — | — | — |

The lint failures at 5.5/5.6 are `@typescript-eslint/parser` 8.x project-service errors ("Parsing error: ... was not found by the project service"), not actual lint violations in our code — that parser claims `typescript: <6.0.0` but its project-service mode misbehaves below 5.7.

Pushed `>=5.7.3 <7` as the new floor in 1293a5f5: that is the lowest TS where every step of our internal toolchain works end-to-end, so we can credibly assert and verify support. Going lower would mean advertising support we cannot exercise in CI.

CLAUDE.md
mike-north: Looked into this and the picture is interesting:

**TS 7 is published as a separate npm package, not a `typescript` major bump.** Per the announcement, the Go rewrite ships under [`@typescript/native-preview@beta`](https://www.npmjs.com/package/@typescript/native-preview) (binary `tsgo`), and it will *eventually* take over the `typescript` package — but not yet. That means our existing matrix logic clamping to `<7` on the `typescript` package is correct and unaffected by the beta — the `typescript@beta` and `typescript@next` dist-tags are still 6.x today.

**The `@typescript/typescript6` compat package** exists explicitly for tools like `typescript-eslint` (and us) that consume the JS API. The recommended pattern for projects on TS 7 is to alias:

```json
"devDependencies": {
  "typescript": "npm:@typescript/typescript6@^6.0.0",
  "@typescript/native-preview": "^7.0.0-beta"
}
```

so `tsgo` runs the new compiler while anything reaching for `import * as ts from "typescript"` still gets the JS API surface.

**Local exploration:** I temporarily added `@typescript/native-preview@beta` and ran `tsgo --noEmit` on a few packages. With deps built, `@formspec/dsl` and `@formspec/analysis` typecheck cleanly. Some packages hit issues — `@types/chai` + `vitest` produce a `Duplicate identifier "containSubset"` under `tsgo` that does not fire under `tsc`, which suggests `tsgo` has stricter declaration-merging rules. That is outside our code but it would block a naive matrix row.

**Recommendation:** track this as a follow-up issue rather than expanding this PR. Adding a `tsgo` matrix row requires:

1. A new dependency (`@typescript/native-preview@beta`) — separate package, not a `pnpm.overrides` flip.
2. A different binary invocation in CI (`pnpm exec tsgo` vs `pnpm exec tsc`).
3. Deciding what `tsgo` exercises: build (it can emit), typecheck (works today modulo `@types/chai`), test (untested — vitest may have its own concerns).
4. Possibly aliasing `typescript` to `@typescript/typescript6` so `@formspec/eslint-plugin`'s `import * as ts` still resolves at runtime when running tests.

Want me to file that as a follow-up issue under epic #424 once this PR lands?

PR #472, opened 2026-04-27
eslint-plugin: fix branded primitive tag checks via analysis capabilities

aidencurtis:
## Motivation

`tag-type-check` had drifted from the canonical type-semantic logic in `@formspec/analysis`. The rule was translating tag capabilities back through local `FieldTypeCategory` helpers, while `@formspec/analysis` already handled nullish-union stripping and branded intersections uniformly. That divergence showed up as false-positive type-mismatch errors for built-in numeric tags on aliases like `Integer = number & { ... }`, and it left the eslint-plugin helper surface asymmetrical across branded primitive shapes.

## Summary

This branch brings `@formspec/eslint-plugin` back into line with the analysis bounded context and fixes the branded-primitive regressions as a cohesive change:

- `tag-type-check` now consults `@formspec/analysis` semantic capabilities directly instead of mapping them back through local `FieldTypeCategory` buckets.
- Built-in numeric tags such as `@minimum`, `@maximum`, `@exclusiveMinimum`, `@exclusiveMaximum`, and `@multipleOf` no longer raise false `typeMismatch` errors on branded numeric aliases.
- The fix covers branded `number`, branded `bigint`, and unions of branded numeric aliases such as `Integer | PositiveInteger`.
- The exported primitive-classification helpers in `@formspec/eslint-plugin` now treat intersections and nullish unions consistently across `string`, `number`, `bigint`, and `boolean`, so the public helper surface no longer has one-off branded-number behavior.
- Regression coverage now includes both positive and negative cases, proving the rule is correctly scoped rather than merely looser: branded strings still reject numeric tags, branded numbers still reject string-only tags, branded bigints are pinned as fixed, and branded numeric unions are covered.

## Test plan

Focused `@formspec/eslint-plugin` validation:

- `vitest` on `tests/rules/constraint-type-mismatch.test.ts`
- `vitest` on `tests/utils/type-utils.test.ts` and `tests/rules/constraint-type-mismatch.test.ts`
- `eslint` on `packages/eslint-plugin/src/utils/type-utils.ts` and `packages/eslint-plugin/tests/rules/constraint-type-mismatch.test.ts`
- `tsc --noEmit` in `packages/eslint-plugin`
- `prettier --check` on the touched eslint-plugin files

Manual validation also performed on a local project.

### DIFF
--- a/packages/eslint-plugin/src/utils/type-utils.ts
+++ b/packages/eslint-plugin/src/utils/type-utils.ts
@@ -78,6 +78,10 @@ export type FieldTypeCategory =
   | "union"
   | "unknown";
 
+function isIntersectionWithBase(type: ts.Type, predicate: (member: ts.Type) => boolean): boolean {
+  return type.isIntersection() && type.types.some((member) => predicate(member));
+}
+
 /**
  * Checks if a TypeScript type is a string type.
  *
@@ -94,29 +98,65 @@ export function isStringType(
   checker: ts.TypeChecker,
   seen = new Set<ts.Type>()
 ): boolean {
-  if (seen.has(type)) {
+  const stripped = stripNullishFromUnion(type);
+  if (seen.has(stripped)) {
     return false;
   }
-  seen.add(type);
+  seen.add(stripped);
 
   // Check for string primitive
-  if (type.flags & ts.TypeFlags.String) {
+  if (stripped.flags & ts.TypeFlags.String) {
     return true;
   }
 
   // Check for string literal
-  if (type.flags & ts.TypeFlags.StringLiteral) {
+  if (stripped.flags & ts.TypeFlags.StringLiteral) {
     return true;
   }
 
   // Check for union of string literals (e.g., "a" | "b")
-  if (type.isUnion()) {
-    return type.types.every((t) => isStringType(t, checker, seen));
+  if (stripped.isUnion()) {
+    return stripped.types.every((t) => isStringType(t, checker, seen));
   }
 
-  const baseConstraint = checker.getBaseConstraintOfType(type);
-  if (baseConstraint !== undefined && baseConstraint !== type) {
+  if (isIntersectionWithBase(stripped, (t) => isStringType(t, checker, seen))) {
+    return true;
+  }
+
+  const baseConstraint = checker.getBaseConstraintOfType(stripped);
+  if (baseConstraint !== undefined && baseConstraint !== stripped) {
     return isStringType(baseConstraint, checker, seen);
+  }
+
+  return false;
+}
+
+function isNumberTypeInner(type: ts.Type, checker: ts.TypeChecker, seen: Set<ts.Type>): boolean {
+  const stripped = stripNullishFromUnion(type);
+  if (seen.has(stripped)) {
+    return false;
+  }
+  seen.add(stripped);
+
+  if (stripped.flags & ts.TypeFlags.Number) {
+    return true;
+  }
+
+  if (stripped.flags & ts.TypeFlags.NumberLiteral) {
+    return true;
+  }
+
+  if (stripped.isUnion()) {
+    return stripped.types.every((t) => isNumberTypeInner(t, checker, seen));
+  }
+
+  if (isIntersectionWithBase(stripped, (t) => isNumberTypeInner(t, checker, seen))) {
+    return true;
+  }
+
+  const baseConstraint = checker.getBaseConstraintOfType(stripped);
+  if (baseConstraint !== undefined && baseConstraint !== stripped) {
+    return isNumberTypeInner(baseConstraint, checker, seen);
   }
 
   return false;
@@ -136,26 +176,29 @@ export function isStringType(
  * @public
  */
 export function isNumberType(type: ts.Type, checker: ts.TypeChecker): boolean {
-  // Check for number primitive
-  if (type.flags & ts.TypeFlags.Number) {
+  return isNumberTypeInner(type, checker, new Set<ts.Type>());
+}
+
+function isBigIntTypeInner(type: ts.Type, seen: Set<ts.Type>): boolean {
+  const stripped = stripNullishFromUnion(type);
+  if (seen.has(stripped)) {
+    return false;
+  }
+  seen.add(stripped);
+
+  if (stripped.flags & ts.TypeFlags.BigInt) {
     return true;
   }
 
-  // Check for number literal
-  if (type.flags & ts.TypeFlags.NumberLiteral) {
+  if (stripped.flags & ts.TypeFlags.BigIntLiteral) {
     return true;
   }
 
-  // Check for union of number literals
-  if (type.isUnion()) {
-    return type.types.every((t) => isNumberType(t, checker));
+  if (stripped.isUnion()) {
+    return stripped.types.every((t) => isBigIntTypeInner(t, seen));
   }
 
-  if (type.isIntersection()) {
-    return type.types.some((t) => isNumberType(t, checker));
-  }
-
-  return false;
+  return isIntersectionWithBase(stripped, (t) => isBigIntTypeInner(t, seen));
 }
 
 /**
@@ -164,16 +207,35 @@ export function isNumberType(type: ts.Type, checker: ts.TypeChecker): boolean {
  * @public
  */
 export function isBigIntType(type: ts.Type): boolean {
-  if (type.flags & ts.TypeFlags.BigInt) {
+  return isBigIntTypeInner(type, new Set<ts.Type>());
+}
+
+function isBooleanTypeInner(type: ts.Type, checker: ts.TypeChecker, seen: Set<ts.Type>): boolean {
+  const stripped = stripNullishFromUnion(type);
+  if (seen.has(stripped)) {
+    return false;
+  }
+  seen.add(stripped);
+
+  if (stripped.flags & ts.TypeFlags.Boolean) {
     return true;
   }
 
-  if (type.flags & ts.TypeFlags.BigIntLiteral) {
+  if (stripped.flags & ts.TypeFlags.BooleanLiteral) {
     return true;
   }
 
-  if (type.isUnion()) {
-    return type.types.every((t) => isBigIntType(t));
+  if (stripped.isUnion()) {
+    return stripped.types.every((t) => isBooleanTypeInner(t, checker, seen));
+  }
+
+  if (isIntersectionWithBase(stripped, (t) => isBooleanTypeInner(t, checker, seen))) {
+    return true;
+  }
+
+  const baseConstraint = checker.getBaseConstraintOfType(stripped);
+  if (baseConstraint !== undefined && baseConstraint !== stripped) {
+    return isBooleanTypeInner(baseConstraint, checker, seen);
   }
 
   return false;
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin/src/utils/type-utils.ts
```diff
@@ -149,6 +151,10 @@ export function isNumberType(type: ts.Type, checker: ts.TypeChecker): boolean {
     return type.types.every((t) => isNumberType(t, checker));
   }
 
+  if (type.isIntersection()) {
+    return type.types.some((t) => isNumberType(t, checker));
+  }
```
mike-north: **Architectural concern: this duplicates a canonical helper that already lives in `@formspec/analysis`.**

`packages/analysis/src/ts-binding.ts` already exposes:

```ts
function isIntersectionWithBase(type, predicate) {
  return type.isIntersection() && type.types.some((member) => predicate(member));
}

function isNumberLike(type) {
  // handles number + number-literal + bigint + bigint-literal
  // strips nullish unions, recurses into unions (every) and intersections (some)
}
```

…plus matching `isStringLike`, `isBooleanLike`, `isNullLike`, and a public capability API (`getTypeSemanticCapabilities` / `hasTypeSemanticCapability`) that already returns `"numeric-comparable"` for branded number-and-bigint intersections.

Per `BOUNDED_CONTEXTS.md`, `StaticAnalysisContext` (`@formspec/analysis`) is the bounded context that owns "TSDoc tag parsing, constraint extraction… the TypeScript-AST work every IDE/lint tool would otherwise re-implement." This file (`@formspec/eslint-plugin`/`type-utils.ts`) is now drifting from that canonical helper — the analysis package does the right thing for four primitives plus null/json/array, while this package now does it for one.

`tag-type-check.ts` already imports `SemanticCapability` from `@formspec/analysis/internal` (line 16) and even maps it back to `FieldTypeCategory` to consult these local helpers. Routing categorization through `hasTypeSemanticCapability(type, checker, "numeric-comparable")` instead of the local `isNumberType`/`getFieldTypeCategory` would:

1. Fix branded `Integer`/`PositiveInteger` (the goal of this PR).
2. Fix branded `string` brands (e.g. `Email = string & { readonly [_emailBrand]: true }`).
3. Fix branded `bigint` brands — closing the gap explicitly called out in the PR description's addendum.
4. Keep this rule honest about who owns type categorization.

If you do want to keep this fix local for now, the strict minimum is to apply the same intersection treatment symmetrically across `isStringType`, `isBigIntType`, and `isBooleanType`. Today the rule will silently misclassify branded strings and branded bigints with the exact same root cause this PR identifies for numbers.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/eslint-plugin/src/utils/type-utils.ts
```diff
@@ -149,6 +151,10 @@ export function isNumberType(type: ts.Type, checker: ts.TypeChecker): boolean {
     return type.types.every((t) => isNumberType(t, checker));
   }
 
+  if (type.isIntersection()) {
+    return type.types.some((t) => isNumberType(t, checker));
```
mike-north: Minor: `isStringType` (line 92) defends against cyclic types with a `seen: Set<ts.Type>` accumulator, but `isNumberType` does not — and this new line recurses through `isNumberType` again. Pathological types (e.g. `type T = number & { self: T }`) would terminate today only because the recursive case eventually hits a non-intersection numeric leaf, but the asymmetry is worth either matching (`seen` set in both) or replacing entirely with the analysis-package helper, which centralizes this concern.

Also, `getBaseConstraintOfType` is sometimes useful as a final fallback in `isStringType` but isn't applied here — another small point of divergence between the two helpers.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/eslint-plugin/tests/rules/constraint-type-mismatch.test.ts
```diff
@@ -170,6 +170,29 @@ ruleTester.run("tag-type-check", tagTypeCheck, {
...
+        };
+
+        export interface MyFields {
+          /** @minimum 0 @maximum 100 */
+          count: Integer;
+
+          /** @exclusiveMinimum 0 @exclusiveMaximum 1000 @multipleOf 5 */
+          batchSize: PositiveInteger;
+        }
+      `,
```
mike-north, line 194: **Test coverage feedback — the new fixture only exercises positive cases.** A regression test for a type-classification rule should also prove the rule still rejects what it should reject; otherwise we can't tell whether the helper was loosened too much. Concretely, the following are missing and would meaningfully strengthen the safety net:

1. **Negative — branded string still rejects numeric tags.** Confirms `isNumberType` doesn't accidentally match branded strings (it shouldn't, because `string & number = never`, but a test pins this behavior):
   ```ts
   export type Email = string & { readonly [__emailBrand]: true };
   /** @minimum 0 */ // expect: typeMismatch
   address: Email;
   ```
2. **Negative — branded number still rejects string tags.** Pins that this PR didn't broaden too far:
   ```ts
   /** @minLength 1 */ // expect: typeMismatch
   count: Integer;
   ```
3. **Documents the bigint gap.** The PR description's addendum acknowledges branded bigint (`bigint & { __brand }`) is unfixed. Either fix it (one line if you adopt the analysis helpers — see the inline comment on `type-utils.ts`) or land a test that captures the current behavior so the gap doesn't regress silently:
   ```ts
   export type Int64 = bigint & { readonly [__int64Brand]: true };
   /** @minimum 0 */ amount: Int64;  // current: false-positive
   ```
4. **A union of branded numerics** (`Integer | PositiveInteger`) — exercises that the union branch composes correctly with the new intersection branch. Since `every` is used inside unions, both members must be classified as numeric for the rule to pass; this is a non-trivial interaction worth pinning.

Concerns 1 and 2 are the ones I would consider blocking — they're the "specificity" tests that demonstrate the fix is correctly scoped, not just "passes for the new shape."

---
_Generated by [Claude Code](https://claude.ai/code)_

mike-north (commented):
Thanks for digging into this — the false positive is real, the diagnosis (branded intersection not unwrapped by `isNumberType`) is correct, and the direction is well-aligned with PP3 ("constraint model mirrors TypeScript's type theory"), S4 ("type determines applicable constraints"), and S7 ("embrace TypeScript's full expressiveness"). The changeset is correctly scoped as a `patch` to `@formspec/eslint-plugin`, and reaching for `ts.TypeFlags.X` enum members rather than literals stays clean of the TS 5↔6 numeric-flag pitfall this file already warns about.

That said, I'd push back on **where** this fix lives and **how broadly** it's applied before merging. Three concerns, in roughly priority order:

### 1. The canonical implementation already exists in `@formspec/analysis` and should be the single source of truth

`packages/analysis/src/ts-binding.ts` already exposes `isIntersectionWithBase` plus matching `isNumberLike`, `isStringLike`, `isBooleanLike`, `isNullLike`, and a public capability API (`getTypeSemanticCapabilities` / `hasTypeSemanticCapability`). All four primitive helpers there already strip nullish unions and unwrap intersections — i.e. the analysis package solved this problem uniformly some time ago.

Per `BOUNDED_CONTEXTS.md`, `StaticAnalysisContext` (`@formspec/analysis`) is the bounded context that owns _"TSDoc tag parsing, constraint extraction… the TypeScript-AST work every IDE/lint tool would otherwise re-implement."_ `EslintContext` is a downstream consumer. The `tag-type-check.ts` rule already imports `SemanticCapability` from `@formspec/analysis/internal` (line 16) and maps each capability back to a `FieldTypeCategory` produced by the local `isNumberType`/`getFieldTypeCategory` — that's a long way around when `hasTypeSemanticCapability(type, checker, "numeric-comparable")` would answer the question directly.

Routing through the analysis-package helper would (a) fix branded `Integer`/`PositiveInteger`, (b) fix branded strings (`Email = string & { __brand }`) which are silently broken today by the same root cause, (c) close the **branded-bigint gap** the addendum calls out, and (d) restore the bounded-context boundary that's eroded as `type-utils.ts` has accumulated independent logic.

### 2. If the fix has to stay local, apply it symmetrically

The PR fixes `isNumberType` only. The same false-positive class exists in `isStringType`, `isBigIntType`, and `isBooleanType` — every primitive helper in this file. The author's addendum already flags bigint; branded strings and branded booleans have the same bug. Patching only one of four creates an asymmetry that's hard to remember and easy to regress when the next branded primitive shows up. Either fix all four here, or — preferred — adopt the analysis-package helper that already does so.

### 3. Tests demonstrate the fix works, but not that it's correctly scoped

Only positive cases were added. Without negative regression tests:

- A branded string with `@minimum` should still fail (specificity check — `string & number = never`, but pin the behavior).
- A branded number with `@minLength` should still fail (proves we didn't broaden numeric classification too far).
- Branded bigint either is fixed or has a test that documents the gap so it doesn't silently regress.
- A union of branded numerics (`Integer | PositiveInteger`) — pins the union+intersection composition path.

I left specifics on the test file. Concerns 1 and 2 are interlinked — if you consume the analysis helpers, the symmetry concern dissolves and the bigint addendum is closed in the same change. I'd be supportive of either approach; my preference is the analysis-package route because it's the architecture this codebase has set up to make exactly this category of bug fix one-line and uniform.

(Specs consulted: `BOUNDED_CONTEXTS.md`, `docs/000-principles.md` PP3/S4/S7/PP14, `docs/005-numeric-types.md` §2.2/§3.1 on integer-by-derivation patterns, and the cross-version `ts.TypeFlags` notes at the top of `type-utils.ts`.)

---
_Generated by [Claude Code](https://claude.ai/code)_

mike-north (commented):
Second-pass review using the TypeScript, simplicity/consolidation, code-quality, testing, and lint-rule lenses.

I do not have additional inline findings beyond the existing comments, but those comments are still the right merge gate from this pass:

- Prefer routing tag applicability through the existing `@formspec/analysis` semantic-capability helpers instead of expanding the local ESLint primitive classifier. That is the simpler implementation and better matches the bounded-context split: analysis owns TypeScript semantic classification; the ESLint rule should consume it.
- If the fix stays local, apply the branded-intersection handling symmetrically across number/string/bigint/boolean helpers rather than only `isNumberType`.
- Strengthen the tests with specificity/negative coverage: branded string + numeric tag should still fail, branded number + string tag should still fail, branded bigint should be either fixed or explicitly pinned, and a union of branded numeric aliases should be covered.

Local validation notes: after installing with `pnpm install --frozen-lockfile` in a temporary PR worktree and building `@formspec/core`, `@formspec/analysis`, `@formspec/config`, and `@formspec/build`, I verified:

- `pnpm --filter @formspec/eslint-plugin run test -- tests/rules/constraint-type-mismatch.test.ts` passed: 17 files, 379 tests.
- `pnpm exec eslint packages/eslint-plugin/src/utils/type-utils.ts packages/eslint-plugin/tests/rules/constraint-type-mismatch.test.ts` passed.
- `pnpm --filter @formspec/eslint-plugin run typecheck` passed.

`pnpm exec prettier --check` reports formatting drift in `packages/eslint-plugin/src/utils/type-utils.ts`, but the same drift is already present on `origin/main` in the untouched `isNullableType` block, so I am not counting that as introduced by this PR.

mike-north:
sorry for the draft/un-draft thrash -- just trying to get the "WIP" github check behaving properly. It works now

aidencurtis:
hey @mike-north - no worries. I've updated to address the review feedback!

mike-north (approved):

